PR #15, opened 2026-05-17
perf: accept optional storage in getUsableFallbackAccounts to avoid redundant disk reads

iceteaSA:
## Summary

Adds an optional `existingStorage` parameter to `FallbackAccountManager.getUsableFallbackAccounts()`. Callers that already have storage loaded (e.g., the fetch handler which loads it at the start of every request) can pass it through to skip a redundant `readFile` + `JSON.parse`.

In the hot request path, `loadAccounts()` currently fires up to 3 times per request:
1. In the main fetch handler
2. Inside `tryFallbackAccounts()`
3. Inside `getUsableFallbackAccounts()`

This change eliminates #3 when the caller provides storage.

## API Change

```typescript
// Before
async getUsableFallbackAccounts(): Promise<OAuthAccount[]>

// After (non-breaking)
async getUsableFallbackAccounts(existingStorage?: AccountStorage | null): Promise<OAuthAccount[]>
```

Uses `!== undefined` check (not falsy) because `null` is a valid "no storage file" state.

## Testing

All existing tests pass. No behavior change — only avoids redundant disk I/O.

### DIFF
--- a/packages/pi/src/convert.ts
+++ b/packages/pi/src/convert.ts
@@ -51,6 +51,13 @@ function sanitize(text: string): string {
   return text.replace(/[\uD800-\uDFFF]/g, '\uFFFD')
 }
 
+function sanitizeToolId(id: string): string {
+  if (!id) return 'tool_call_unknown'
+  const sanitized = id.replace(/[^a-zA-Z0-9_-]/g, '_')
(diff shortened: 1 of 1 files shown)


iceteaSA:
Closing — wrong branch mapping, will resubmit.

PR #25, opened 2026-05-17
perf(relay): defer KV state write, add health endpoint, harden POST handler

iceteaSA:
## Summary

Four improvements to the relay worker script:

### 1. Deferred KV state write (~4ms P50 improvement)

`prepareUpstream` previously `await`ed the KV `writeState` call before returning the body. Now the write is fire-and-forget (`.catch(() => {})`) and deferred via `ctx.waitUntil`. Response streaming starts immediately — the state is persisted in the background.

### 2. Heartbeat interval 5s → 15s

Reduces WebSocket keepalive overhead by 3x. The CF Workers runtime keeps connections alive independently; the heartbeat only needs to prevent idle timeouts on intermediate proxies, which typically have 30-60s thresholds.

### 3. GET `/` health endpoint

Returns JSON with relay status and available transports instead of plain "ok":
```json
{"status":"ok","transports":["http","websocket"]}
```

### 4. POST handler try/catch → 502

Wraps the HTTP POST handler in a try/catch that returns structured JSON `502` instead of raw Cloudflare error pages on unexpected failures. Clients can parse the error and fall back to direct API calls.

## Testing

All 214 tests pass. Changes are in the embedded `WORKER_SCRIPT` template string — no client-side behavior changes.

### DIFF
--- a/packages/core/src/relay.ts
+++ b/packages/core/src/relay.ts
@@ -1027,7 +1027,7 @@ async function prepareUpstream(env, payload) {
     return { error: 'hash mismatch', status: 409 }
   }
 
-  await writeState(env, payload.affinity, { body, hash: payload.next_hash, revision: payload.revision })
+  const stateWrite = writeState(env, payload.affinity, { body, hash: payload.next_hash, revision: payload.revision }).catch(() => {})
   console.log(JSON.stringify({
     relay: 'opencode-anthropic-auth',
     transport: 'relay',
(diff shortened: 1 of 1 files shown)


ualtinok:
Thanks a lot for all the PRs, all good fixes and additions. 

If you're using discord, please join us https://discord.gg/F2uWxjGnU

iceteaSA:
> Thanks a lot for all the PRs, all good fixes and additions.
> 
> If you're using discord, please join us https://discord.gg/F2uWxjGnU

A pleasure. Some more coming soon, when I have time to split out the features I have in my frankenstein's monster of a fork. Already joined :+1:

PR #32, opened 2026-05-21
fix(pi): narrow converter guards — trailing assistant, tool/image/surrogate, signed-thinking

iceteaSA:
Narrow, Pi-owned converter hardening for `packages/pi/src/convert.ts`, beyond what was merged in #17. All changes are scoped to concrete invalid shapes the Pi converter can produce — no broad payload "healing".

**Commit 1: strip trailing assistant messages, improve sanitize and error guards**

- **Strip trailing assistant messages** from converted requests to prevent Anthropic prefill 400s on models that reject assistant-message prefill.
- **Unicode regex** (`/gu` flag) for spec-compliant surrogate handling — valid surrogate pairs (emoji / astral characters) are preserved; only lone surrogates are replaced with `\uFFFD`.
- **Simplified `sanitizeToolId`** with a cleaner length check (removes a redundant guard).
- **Error-content guard** injects an `Error` placeholder for `is_error=true` tool results with empty content (`!content`, or an empty array).
- Comprehensive test coverage, including a regression test asserting that valid surrogate pairs survive sanitisation.

**Commit 2: preserve signed thinking verbatim, guard lone surrogates**

- **Send signed thinking blocks byte-identical.** The signature is computed over the original text, so sanitizing would alter it and Anthropic rejects the block as *modified*. Previously the converter sanitized signed thinking unconditionally; now it preserves it verbatim (the normal path for Anthropic-origin reasoning).
- **Lone-surrogate exception only.** A signed block containing a lone (unpaired) UTF-16 surrogate can't keep its signature — sanitizing breaks it, and the raw surrogate is invalid UTF-8 (400). `hasLoneSurrogate()` detects exactly that case, and only then is the signature dropped and the text sanitized.
- Tests cover verbatim signed-thinking preservation and the lone-surrogate downgrade.

### Scope note (per review)

This PR intentionally **does not** include the broad "downgrade all historical signed thinking to text" healing. That approach can avoid 400s but discards valid signed/redacted reasoning when the request shape is otherwise correct. The signed-thinking **reordering** problem is better fixed at the OpenCode layer in anomalyco/opencode#30182. For Pi, the converter should avoid broad destructive healing unless a concrete Pi-generated invalid shape requires it — none currently does, so it is omitted here.

This PR hardens the Pi converter (`packages/pi/src/convert.ts`) with four targeted fixes: stripping trailing assistant messages to avoid Anthropic prefill 400s, preserving signed thinking blocks verbatim (previously they were sanitized, which broke the signature), downgrading signed blocks that contain lone surrogates, and tightening the empty-content guard for `is_error=true` tool results.

- **Signed-thinking verbatim pass-through** (`convert.ts` lines 149-168): the old code called `sanitize()` on signed thinking text, which mutated the content and invalidated the signature. The new code sends the text byte-for-byte identical, and only falls back to sanitized text when `hasLoneSurrogate()` detects an unpaired surrogate that would cause a UTF-8 400 anyway.
- **Trailing-assistant stripping** (`buildAnthropicRequest`, lines 306-312): a `while` loop removes any assistant messages left at the tail of the converted array before the request is sent, covering models that reject assistant-role prefill.
- **Test suite added** (`convert.test.ts`, 428 lines): covers all new guards including surrogate sanitisation, signed-thinking preservation, error tool-result placeholder injection, and the trailing-assistant strip.
</details>

Safe to merge — all changes are narrowly scoped to concrete invalid shapes the Pi converter produces, and the new test suite exercises every guard added.

The signed-thinking verbatim pass-through correctly identifies that sanitizing a signed block breaks its signature, and the lone-surrogate fallback covers the only edge case where verbatim sending would itself cause a 400. The trailing-assistant strip is a simple while-loop applied after conversion, which cannot affect any mid-conversation message. The error-content guard correctly handles the full string | Array union returned by convertTextAndImages. Test coverage is thorough and the emoji/astral-character preservation regression test is present.

No files require special attention.

| Filename | Overview |
|----------|----------|
| packages/pi/src/convert.ts | Core converter hardening: verbatim signed-thinking pass-through, lone-surrogate detection, trailing-assistant strip, and simplified empty-error-content guard — all changes are logically correct and well-commented. |
| packages/pi/src/tests/convert.test.ts | New test file with comprehensive coverage of all changed behaviours; assertions are correct and the emoji/astral-character preservation case is included (lines 219-227). |
| packages/pi/package.json | Adds `"test": "bun test src/tests"` script; no other changes. |

</details>

</details>

<details><summary><h3>Flowchart</h3></summary>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[thinking block] --> B{thinkingSignature set?}
    B -- No --> E[sanitize text → emit as text block]
    B -- Yes --> C{hasLoneSurrogate?}
    C -- No --> D[emit verbatim thinking block\nwith original text + signature]
    C -- Yes --> F[drop signature\nsanitize text → emit as text block]

    G[tool_result content] --> H{isError = true?}
    H -- No --> I[emit content as-is]
    H -- Yes --> J{content empty/blank?}
    J -- No --> I
    J -- Yes --> K[inject placeholder: Error]

    L[converted messages array] --> M{trailing message role = assistant?}
    M -- Yes --> N[pop and repeat]
    M -- No --> O[send to Anthropic]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(pi): preserve signed thinking verbat..."](https://github.com/cortexkit/anthropic-auth/commit/3bcd3e9c0609e8bc5333c7826bb0d176de7a08b2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35109361)</sub>

### DIFF
--- a/packages/pi/package.json
+++ b/packages/pi/package.json
@@ -32,6 +32,7 @@
     "build": "rm -rf dist && tsc -p tsconfig.build.json",
     "typecheck": "tsc",
     "types": "tsc",
+    "test": "bun test src/tests",
     "prepublishOnly": "bun run build"
   },
   "dependencies": {
--- a/packages/pi/src/convert.ts
+++ b/packages/pi/src/convert.ts
@@ -48,14 +48,18 @@ export type AnthropicRequestBody = {
 }
 
 function sanitize(text: string): string {
-  return text.replace(/[\uD800-\uDFFF]/g, '\uFFFD')
+  return text.replace(/[\uD800-\uDFFF]/gu, '\uFFFD')
 }
 
+/**
+ * Sanitize a tool-call ID to match Anthropic's `^[a-zA-Z0-9_-]+$` pattern.
+ * Cross-provider IDs (e.g. OpenAI Codex `call_xxx|fc_xxx`) contain characters
+ * Anthropic rejects. Deterministic — same input always yields the same output.
+ */
 function sanitizeToolId(id: string): string {
   if (!id) return 'tool_call_unknown'
-  const sanitized = id.replace(/[^a-zA-Z0-9_-]/g, '_')
-  if (!sanitized) return 'tool_call_unknown'
-  return sanitized.slice(0, 256)
+  const cleaned = id.replace(/[^a-zA-Z0-9_-]/g, '_')
+  return cleaned.length > 256 ? cleaned.slice(0, 256) : cleaned
 }
 
 function toClaudeCodeToolName(name: string): string {
@@ -157,17 +161,20 @@ function convertMessages(
     if (message.role === 'toolResult') {
       const toolResult = message as ToolResultMessage
       const toolResults: Array<Record<string, unknown>> = []
-      const firstContent = convertTextAndImages(toolResult.content)
-      const firstContentArr = Array.isArray(firstContent)
-        ? firstContent
-        : [{ type: 'text', text: firstContent }]
-      if (toolResult.isError && firstContentArr.length === 0) {
-        firstContentArr.push({ type: 'text', text: 'Error' })
+      let content = convertTextAndImages(toolResult.content)
+      // Anthropic rejects tool_result with is_error=true but empty content
+      if (
+        toolResult.isError &&
+        (!content ||
+          (Array.isArray(content) && content.length === 0) ||
+          content === '')
+      ) {
+        content = [{ type: 'text', text: 'Error' }]
       }
       toolResults.push({
         type: 'tool_result',
         tool_use_id: sanitizeToolId(toolResult.toolCallId),
-        content: firstContentArr,
+        content: content,
         is_error: toolResult.isError,
       })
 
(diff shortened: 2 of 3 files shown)


ualtinok:
Reviewed this in context with OpenCode PR https://github.com/anomalyco/opencode/pull/30182 and the related OpenCode-side healing PRs here (#49, #51).

My read:

- The Pi conversion tests and the narrower conversion hardening are useful: trailing assistant stripping, empty `is_error` tool result handling, tool ID/image/surrogate guards are all reasonable Pi-owned converter fixes.
- The historical signed-thinking downgrade should not be merged as-is. It is the same broad “heal the final payload by converting signed thinking to text” approach as #49. That can avoid Anthropic 400s, but it also discards valid signed/redacted thinking that should be preserved when the converter/request shape is correct.
- For OpenCode, PR #30182 is the better layer for the signed-thinking issue because it prevents OpenCode’s Anthropic reorder from moving signed/redacted reasoning in the first place. For Pi, OpenCode cannot fix the converter, but the Pi converter should still avoid broad destructive healing unless a concrete Pi-generated invalid shape requires it.

Recommendation: split this PR. Keep the tests and the narrow Pi conversion guards, including trailing assistant stripping. Drop or separately rework the historical signed-thinking downgrade into a much narrower fix backed by a concrete Pi repro.

iceteaSA:
> Recommendation: split this PR. Keep the tests and the narrow Pi conversion guards, including trailing assistant stripping. Drop or separately rework the historical signed-thinking downgrade into a much narrower fix backed by a concrete Pi repro.

Dropped it

PR #33, opened 2026-05-21
feat(relay): plan-based gating, KV error logging, and request logging

iceteaSA:
## Summary
Relay Worker improvements + a CLI deploy fix so the worker's plan gating actually works.

### Worker (`relay.ts` `WORKER_SCRIPT`)
- **Plan-based gating** via the `RELAY_PLAN` env binding: `paid` enables WebSocket transport + structured logging; otherwise the worker stays HTTP-only (WS upgrades return 403) to fit the free-tier CPU budget.
- **KV error logging** with a 7-day TTL.
- **Request logging** (paid plan only).

### CLI deploy (`cli.ts`)
- `relaySetup` now selects the plan — `RELAY_PLAN`/`CLOUDFLARE_PLAN` env, else prompt (**default `free`**, the safe option for any account) — and passes it through.
- `uploadRelayWorker` sets the `RELAY_PLAN` binding on every deploy. A worker `PUT` replaces all bindings, so without this the worker silently reverted to free-plan behaviour on redeploy.
- Relay `transport` now matches the plan (`websocket` for paid, `http` for free).

### Tests
- `cli.test.ts` asserts the relay-setup deploy carries the `RELAY_PLAN` binding and selects the matching transport.
- `relay-worker-miniflare.test.ts` covers the gated worker behaviour.

### DIFF
--- a/packages/core/src/relay.ts
+++ b/packages/core/src/relay.ts
@@ -1191,6 +1191,11 @@ export async function sendViaRelay(options: {
 }
 
 export const WORKER_SCRIPT = `
+function getPlanConfig(env) {
+  const paid = (env.RELAY_PLAN || '').toLowerCase() === 'paid'
+  return { paid, allowWebSocket: paid, logRequests: paid }
+}
+
 async function hashBody(body) {
   const bytes = new TextEncoder().encode(body)
   const digest = await crypto.subtle.digest('SHA-256', bytes)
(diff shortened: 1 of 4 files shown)


ualtinok:
Closing this because the relay implementation has moved on and the remaining error-logging design stores upstream response bodies and headers in KV and logs response previews. We do not want that privacy exposure in the relay diagnostics path. Thanks for the work on the original relay gating and diagnostics proposal.

PR #34, opened 2026-05-21
feat(core): add QuotaManager with exponential backoff and persistence

iceteaSA:
Unified quota cache and API gateway for main + fallback quota state. All consumers share one `QuotaManager` instance for consistent caching, inflight deduplication, and 429 backoff.

Reworked to address review feedback — **QuotaManager is now the single quota path**, not a second path beside the existing fallback logic.

## What changed in this rework

**One quota path (was: two)**
- `FallbackAccountManager.refreshAccountQuota()` now routes through `QuotaManager.refreshFallback()` (inflight dedup + 429 backoff) instead of calling `fetchOAuthQuotaSnapshot()` directly. Direct fetch remains only as a fallback when no QuotaManager is wired (isolation/tests).
- Fallback selection and background refresh therefore share the same gated path as everything else.

**`/claude-quota` forces real requests and persists (regression fixed)**
- Main: forces `refreshMain()` (no cache short-circuit).
- Fallbacks: `refreshQuotaForAllAccounts({ force: true })` bypasses the staleness skip, refreshes every enabled account through the unified path, **persists** refreshed snapshots to `anthropic-auth.json`, and clears stale `lastQuotaRefreshError`.
- 429 backoff is still respected — when the API is backed off it surfaces rate-limited and shows the last cached numbers rather than hammering.

**Request-path refresh trigger wired (was: dead config)**
- `needsRefresh(sessionRequestCount)` now drives an every-N background refresh for **main** on the request path (`quota.refreshEveryNRequests`), alongside the existing time-staleness trigger.
- **Fallbacks refresh proactively too**, matching main: the existing background timer ticks `refreshQuotaForDueAccounts()`, and the **currently-active fallback route** refreshes on the same every-N cadence at its `markUsed()` point. Idle fallbacks are never hit on the request path — they rely on the background timer.

**Backoff hardening (cubic)**
- `getUsableFallbackAccounts()` skips the request-time refresh when an account's quota API is backed off (`quotaBackoffActive`), mirroring the background path — no more hammering a failing quota API on every request.
- A **401 no longer arms quota backoff**: it is an auth/token-expiry signal, so the caller refreshes the token and retries; backing off would have blocked that retry and every other account.

**Fail-closed killswitch moved out of the base**
- The main-account synthetic-429 fail-closed (block when quota API backed off + `failClosedOnUnknownQuota`) is **removed from this branch**. It is a killswitch behavior and moves to `feat/killswitch`. This base is a pure routing optimization that never blocks the main request.

## Unchanged base features
- Inflight deduplication; exponential backoff (60s–15min) for 429/5xx; persists main quota + backoff state via callbacks; cross-process file-lock guard; seeds from persisted storage on construction.

## Review items addressed
1. ✅ Fallback quota refresh no longer bypasses QuotaManager — unified path.
2. ✅ `/claude-quota` persists refreshed fallback quota again (+ clears stale errors).
3. ✅ `refreshEveryNRequests` / `needsRefresh` are now wired (main request-path + active-route fallback).
4. ✅ Fail-closed killswitch moved to `feat/killswitch` — base no longer changes main-request behavior.
5. ✅ (cubic) `getUsableFallbackAccounts` now checks per-account `quotaBackoffActive` before refresh.
6. ✅ (cubic) `saveAccounts` calls pass `accountStoragePath` (verified).

Rebased onto current `main`. Tests: 279 opencode pass, typecheck clean, biome clean. New tests cover forced-refresh persistence, the 401 no-backoff behavior, and the active-route every-N trigger.

## Follow-up review fixes

Subsequent AI review found three more issues, now fixed:

- **Per-route backoff scoping (P1).** A single shared `lastApiError` meant a fallback account's quota `429` backed off the **main** account and persisted as `storage.quota.mainLastQuotaApiError` (via `onApiError`), so `refreshMain` would short-circuit and throw without ever trying the main token. Backoff is now scoped: `mainLastApiError` drives `isBackedOff()`/`getLastApiError()`/`refreshMain` and only main failures arm it or call `onApiError`; a per-account `fallbackApiErrors` map drives the new `isFallbackBackedOff(accountId)`, and fallback failures never touch main state (their error is already recorded per-account via `lastQuotaRefreshError`).
- **Token-bound main seed (P2).** The persisted main quota was seeded into the cache without recording which account produced it, so `refreshMain`'s token-change guard had no baseline and a different account's stale quota could be served during backoff. A truncated SHA-256 token fingerprint is now persisted with the main quota and seeded alongside it; `refreshMain` drops the seed before the backoff short-circuit when the live token's fingerprint differs.
- **Request-count cadence (P1, cubic).** `sessionRequestCount` is now incremented for every replayable request instead of only inside the `mainQuotaRoutingEnabled` branch, so the every-N refresh cadence (and the active-route fallback refresh) advances on all routing paths.

Adds regression tests for all three. Full suite: 283 opencode pass, typecheck clean, biome clean.

## Second-round review fixes (request flow)

- **Token-aware cached main read.** The routing path read `quotaManager.getMain()?.quota` before `refreshMain`, bypassing the token-fingerprint check, so a fresh cached snapshot from a previous main account could be used for a new access token. `getMain(accessToken?)` now drops a cached entry bound to a different token (main-account switch) and the routing path passes `auth.access`.
- **Count fallback-first requests.** `sessionRequestCount` incremented after the fallback-first early return, so a successful fallback-first request never advanced the every-N cadence and the served fallback's active-route refresh never fired. The counter now increments right after `replayableRequest` is determined, before any routing/early-return.

Adds regression tests for both. Full suite: 285 opencode pass, typecheck clean, biome clean.

## Third-round review fix (fallback source of truth)

Fallback selection used the QuotaManager cache for the staleness check but evaluated quota policy from the storage `account.quota`, while the active-route every-N refresh updates only the cache. The two could diverge:
- A fresh-but-exhausted cache entry suppressed a refetch yet was ignored by the policy decision → routed from stale passing storage quota.
- A same-id re-login (cleared `account.quota`, stale fresh cache) skipped the fetch then failed closed on missing `account.quota`.

Fix — one source of truth:
- `getUsableFallbackAccounts` evaluates policy via `quotaPolicyAccount()`, overlaying the QuotaManager-cached quota (same source as the staleness check).
- Fallback cache entries are token-bound: `getFallback`/`isFallbackStale` accept the live access token and drop an entry produced by a different token (re-login); `_fetchFallback` records the fingerprint. Selection passes `next.access` so a credential change invalidates the stale entry and refetches.

Adds regression tests for both repros. Full suite: 287 opencode pass, typecheck clean, biome clean.


ualtinok:
I retested this against current `main`. A temp merge is clean, and the focused local checks passed:

- `bun run build`
- `bun run typecheck`
- `bun test packages/opencode/src/tests/quota-manager.test.ts packages/opencode/src/tests/index.test.ts`

I do not think this is ready to merge as-is, though. The main concern is that it introduces `QuotaManager` beside the existing fallback quota path rather than fully replacing it.

Specific issues:

1. **Fallback quota refresh still bypasses `QuotaManager`.**
   `FallbackAccountManager.refreshAccountQuota()` still calls `fetchOAuthQuotaSnapshot()` directly. That means fallback selection/background refresh continue using the old path, while `QuotaManager` is only used for main quota and `/claude-quota` fallback display refreshes.

2. **`/claude-quota` fallback quota persistence appears to regress.**
   Current `main` uses `fallbackManager.refreshQuotaForAllAccounts()`, which persists refreshed fallback quota snapshots and clears stale `lastQuotaRefreshError` when stored quota is still fresh. This PR switches the command to `quotaManager.refreshAllFallbacks()` and then overlays cached quota onto the in-memory storage object for display, but does not save the refreshed fallback quota back to `anthropic-auth.json`.

3. **`refreshEveryNRequests` is added but not wired.**
   `QuotaManager.shouldRefreshOnRequestCount()` / `needsRefresh()` are not called anywhere. Either the request-path trigger should be implemented, or the new config surface should be removed for now.

4. **The main-account fail-closed killswitch is a behavior change.**
   If `quota.failClosedOnUnknownQuota` is explicitly true, there is no cached main quota, and the quota API is backed off, the fetch wrapper returns a synthetic 429 before sending the model request. Current behavior treats main quota checks as a routing optimization and does not block the main request when the quota probe fails. That product behavior should be explicitly decided before landing.

I’d rework this so there is one actual quota path, especially for fallback refresh/selection and `/claude-quota` persistence, then we can re-review.

iceteaSA:
I agree, there are also issues with the new routing not properly refreshing quota as I expect as well on fallbacks, so this definitely needs some more TLC. Will do it along with the TUI sidebar changes, as everything is pretty closely linked. This branch is the base for a lot of my PRs

iceteaSA:
@ualtinok Done

ualtinok:
@cubic-dev-ai review

ualtinok:
Cubic already covered the `sessionRequestCount` / every-N refresh issue, so I won't duplicate that thread.

One additional blocker I found in the updated branch: quota API backoff is still global across main and all fallback accounts, and fallback quota errors are persisted through the main quota error callback.

Source path:

- `QuotaManager` has one shared `lastApiError`.
- `_fetchFallback()` calls `_handleFetchError()` on fallback quota failures.
- `_handleFetchError()` sets that shared `lastApiError` and invokes `onApiError`.
- The OpenCode plugin's `onApiError` persists the error as `storage.quota.mainLastQuotaApiError`.
- Later, `refreshMain()` checks `isBackedOff()` before attempting the main quota fetch. If no main quota is cached, it throws `Quota API rate-limited — try again later` without trying the main token.

I reproduced the behavior directly on this branch:

```text
fallback error: Error: Claude quota check failed: 429 — rate limited
calls after fallback 429: [ "Bearer fallback-token" ]
onApiError count: 1
backed off after fallback 429: true
main after fallback 429: Error: Quota API rate-limited — try again later
calls after main attempt: [ "Bearer fallback-token" ]
```

So a fallback account's quota `429` can suppress main quota checks and persist as a main quota API error. That is not safe for account-level routing decisions; quota/backoff state needs to be scoped by route/account/token, or at minimum fallback quota failures must not update `mainLastQuotaApiError` or block main quota refreshes.

Local validation otherwise passed on a temp merge (`bun run typecheck`, `bun run test`), but I would hold merge until this backoff scoping issue and cubic's comments are resolved.

ualtinok:
@cubic-dev-ai review plz

ualtinok:
Rechecked the latest update. The scoped fallback/main backoff change fixes the issue I reported earlier: fallback quota `429` no longer calls `onApiError`, no longer sets main backoff, and a later main quota fetch can proceed.

I still do not think this is mergeable yet. I found two remaining request-flow bugs that Cubic's latest re-review did not catch.

1. **Token-bound main quota is still read without checking the live token.**

`QuotaManager.refreshMain(accessToken)` now invalidates cached/persisted main quota when `mainTokenFp` differs, but the OpenCode request path reads `quotaManager.getMain()?.quota` first:

```ts
let routingQuota = quotaManager.getMain()?.quota
if (!routingQuota) {
  routingQuota = await quotaManager.refreshMain(auth.access)
}
```

Because `getMain()` takes no access token, this bypasses the token-fingerprint check. If main auth changes while a cached main quota snapshot is still fresh, routing can use the previous account's quota for the new access token.

Runtime repro on this branch:

```text
status: 200
usage calls: 0
request authorizations: [ "Bearer fallback-access" ]
```

That repro seeded `mainQuotaToken` for `old-main-access`, then ran a request with `new-main-access`. The stale cached main quota said main was exhausted, so the plugin routed directly to fallback. It made **zero** `/api/oauth/usage` calls, so the new main token was never checked. The opposite stale-cache shape would send traffic to a newly exhausted main account.

Fix direction: the request path needs a token-aware cache read, e.g. `getMain(auth.access)` / `getMainForToken(auth.access)`, or it should call a QuotaManager method that validates the live token fingerprint before returning a cached main entry.

2. **Successful `fallback-first` requests still do not advance the every-N quota refresh counter.**

The counter was moved out of the `mainQuotaRoutingEnabled` branch, but it still increments after the `fallback-first` early-return path. When fallback-first succeeds, the function returns before this runs:

```ts
if (replayableRequest) sessionRequestCount++
```

Runtime repro with `routing.mode = "fallback-first"` and `quota.refreshEveryNRequests = 1`:

```text
status: 200
usage calls after fallback-first request with refreshEveryNRequests=1: 0
request authorizations: [ "Bearer fallback-access" ]
```

Given `refreshEveryNRequests = 1`, the served fallback should have triggered a quota refresh after the successful request. It did not, because `tryUsableFallbackAccounts()` saw `sessionRequestCount === 0`.

Fix direction: increment the replayable request counter before any fallback-first early return, then use that value for main and fallback active-route refresh checks.

Local validation otherwise still passes on a temp merge (`bun run typecheck`, `bun run test` → `285 pass, 0 fail`), and `aft_inspect` reported no diagnostics/TODOs for the changed core/index files.

ualtinok:
Rechecked the latest update. The two issues from my previous comment are fixed:

- `getMain(auth.access)` now validates the cached main quota against the live token.
- `sessionRequestCount` now increments before the fallback-first early return.

I verified both former repros now pass at runtime:

```text
stale-token usage auths: [ "Bearer new-main-access" ]
stale-token request auths: [ "Bearer new-main-access" ]
fallback-first usage auths: [ "Bearer fallback-access" ]
fallback-first request auths: [ "Bearer fallback-access" ]
```

I found one remaining blocker in the fallback quota path.

`QuotaManager` fallback cache is used for staleness, but fallback routing still evaluates quota policy from the `OAuthAccount` object loaded from storage:

- `getUsableFallbackAccounts()` checks `quotaManager.isFallbackStale(next.id)`.
- If the QuotaManager entry is fresh, it skips `refreshAccountQuota()`.
- Then `accountPassesQuotaPolicy(next, storage)` reads `next.quota`, not `quotaManager.getFallback(next.id)?.quota`.
- The new active-route every-N refresh calls `quotaManager.refreshFallback(account.id, access)` directly, which updates only the in-memory QuotaManager cache, not `account.quota` in storage.

So the QuotaManager cache can suppress a refresh while being ignored for the actual routing decision.

Direct repro on this branch:

```text
QM cached remaining: 0
storage remaining: 100
usable ids despite QM failing quota: [ "fallback-1" ]
```

That simulates an active-route refresh updating QuotaManager to an exhausted quota while storage still has an older passing quota. `getUsableFallbackAccounts()` treats the QuotaManager entry as fresh, does not refetch, then routes using the stale storage quota and returns the fallback as usable.

The inverse is also possible: after same-label re-login clears `account.quota`, a still-running plugin process can have a fresh old QuotaManager cache for the same account id. Then selection skips the quota fetch but fails closed on missing `account.quota`:

```text
is stale before: false
calls: []
usable: []
quota after: undefined
```

Fix direction: fallback selection needs one quota source of truth. Either:

1. make active-route fallback refresh go through a manager path that persists/updates `account.quota`, or
2. make `getUsableFallbackAccounts()` overlay/use the QuotaManager fallback entry for policy decisions, and invalidate/token-bind fallback cache entries when the stored account credentials change.

Until then, the PR can route from stale fallback quota state even though the unified manager has fresher data.

Local validation still passes on the temp merge (`bun run typecheck`, `bun run test` → `287 pass, 0 fail`), so this is a logic/integration gap rather than a test failure.

ualtinok:
Rechecked latest `b8723307`. The previous stale-storage-policy bug is fixed for active-route cache entries: policy now overlays `quotaManager.getFallback(account.id, account.access)` and the exhausted-QM/stale-storage repro now returns no usable account.

There is one remaining token-binding hole: fallback quota entries seeded from persisted storage are still tokenless.

Source path:

- `QuotaManager.getFallback(accountId, accessToken)` invalidates only if `fallbackTokenFps` has a fingerprint and it differs.
- `_fetchFallback()` records `fallbackTokenFps`, so active-route refresh entries are token-bound.
- But `QuotaManager.setFallback(accountId, entry)` does not accept/set an access token.
- `seedFallbacksFromAccounts()` and `FallbackAccountManager.seedFallbackQuota()` both seed persisted `account.quota` through `setFallback()` without binding it to `account.access`.
- After same-label re-login, disk `account.quota` is cleared, but an already-running plugin process can still hold a fresh tokenless QM entry from the old credentials. Because it has no fingerprint, `getFallback(id, newAccess)` treats it as valid.

Plugin-level repro on this branch:

```text
usage auths: []
model auths: [ "Bearer old-fallback-access", "Bearer new-fallback-access" ]
```

That repro:

1. starts the plugin with `same-label` / `old-fallback-access` and fresh persisted quota, causing QM to seed a tokenless fallback cache entry;
2. simulates CLI re-login in the same still-running plugin process by replacing the sidecar account with `same-label` / `new-fallback-access` and no `account.quota`;
3. sends another fallback-first request.

The second request routes with the new token but makes **zero** `/api/oauth/usage` calls, because the old tokenless QM quota cache is still considered fresh.

Fix direction: token-bind persisted fallback seeds too. For example, make `setFallback` accept an optional access token and have `seedFallbacksFromAccounts()` / `seedFallbackQuota()` pass `account.access`; or have token-aware `getFallback(id, accessToken)` treat un-fingerprinted entries as invalid. The first option is less disruptive because display/tokenless reads can still work while request-time reads become safe.

Validation still passes on the temp merge:

- `bun run typecheck`
- `bun run test` → `289 pass, 0 fail`
- `aft_inspect` on changed core/index files: no diagnostics/TODOs

PR #36, opened 2026-05-21
feat(opencode): show quota usage toast after quota refresh

iceteaSA:
> Rebased onto `upstream/main` (v1.5.0) now that #34 (QuotaManager) is merged — this branch no longer depends on `feat/quota-manager`.

Displays quota-usage bar notifications via `client.tui.showToast` after quota data is refreshed.

- Shows main and enabled fallback account usage with visual bars (`█░`).
- Padded percentage and reset time per account; `formatResetIn` handles sub-1-minute (`<1m`), "now", and invalid-date inputs.
- Toast variant reflects severity: info (<70%), warning (≥70%), error (≥90%).
- Clamped bar rendering prevents `RangeError` on >100% utilization.
- Reads main/fallback quota from the shared, **token-aware** QuotaManager cache (not the request-start snapshot).
- The "active" account label mirrors actual routing: the first fallback that passes quota policy, otherwise **main** — never a failing fallback.
- Fires on a synchronous cold-start cache miss and on the resolved `.then()` of each periodic background refresh; never blocks the request.

Single file change: `packages/opencode/src/index.ts`.

---
## Summary by cubic
Shows a quota usage toast after quota refresh, matching the TUI sidebar. Keeps the sidebar correct during async refreshes by preserving the active account and re-rendering when background fallback storage changes via `@cortexkit/anthropic-auth-core`.

- **New Features**
  - Quota toast via `client.tui.showToast` after cold-start `refreshMain` and background refresh; shows main + enabled fallbacks with shared-width `█░` bars, padded %, 5h/7d, “resets <1m/now”, and active/idle status; clamps >100%; severity by max usage; reads token-aware data from the shared `QuotaManager` cache and marks the first account that passes policy.

- **Bug Fixes**
  - Preserve the active fallback in the sidebar after async main refresh by remembering the last routing and only refreshing numbers.
  - Auto-refresh the sidebar when background fallback storage changes (token refresh, quota update, error) by wiring `onFallbackStorageChanged` in `FallbackAccountManager`.

<sup>Written for commit 4b7a88c3a3352c3244e5646b797fffcb32852990. Summary will update on new commits.</sup>

[...]

PR #35, opened 2026-05-21
feat(opencode): killswitch — block requests when quota drops below threshold

iceteaSA:
> Rebased onto `upstream/main` (v1.5.0) now that #34 (QuotaManager) is merged — this branch no longer depends on `feat/quota-manager`.

Per-account request blocking when remaining quota drops below configurable thresholds. Returns a synthetic 429 only when the request genuinely cannot be routed anywhere.

**Features:**
- Per-account threshold overrides (5h and 7d windows), persisted to the sidecar `killswitch` block.
- Eager quota refresh on the first request, with a **fresh re-read before evaluation** so the killswitch fires correctly even on the very first request (the pre-refresh snapshot is `null`).
- Skip-main routing when main is killed (try surviving fallbacks). A synthetic **429** is returned only when main is killed **and** every fallback is killed **or** the request body is non-replayable (so it cannot be routed to a fallback). A request that is merely below the soft routing threshold still falls through to main.
- `killswitchPassesPolicy` evaluates **all present** quota windows — a present below-threshold window still blocks even when the other window is missing (`failClosed=false`).
- All killswitch / fallback decisions **and** the `Retry-After` computation read the fresh, **token-aware** QuotaManager cache, never the stale request-start storage snapshot.
- `Retry-After` header set to the earliest quota reset across the unroutable accounts.
- `/claude-killswitch` slash command (`on` / `off` / `set` / `status`) for runtime management.

**Files:**
- `packages/core/src/killswitch.ts` — new
- `packages/core/src/accounts.ts` — killswitch types and policy functions (shared `DEFAULT_KILLSWITCH_THRESHOLDS`)
- `packages/opencode/src/index.ts` — killswitch fetch-gate + command registration
- `packages/opencode/src/tests/killswitch.test.ts` + `index.test.ts` fetch-gate tests
- `README.md` + `packages/opencode/README.md` — killswitch documentation

---
## Summary by cubic
Adds a per‑account killswitch and a unified quota cache in `@cortexkit/anthropic-auth-core` so OpenCode blocks low‑quota requests and returns 429 with Retry‑After when unroutable or when the quota API is backed off. Adds `/claude-killswitch`, enforces the killswitch across all routing paths, and keeps the sidebar correct after reroutes and background updates.

- **New Features**
[...]

PR #38, opened 2026-05-21
feat(opencode): add killswitch indicators to quota toast

iceteaSA:
## Summary
Adds killswitch indicators to the quota usage toast.

Builds on the merged killswitch (#35) and quota toast (#36). Each account line in the toast now shows a `blocked` status (instead of `active`/`idle`) when the killswitch has removed it from the routing pool, so the toast matches the routing reality at a glance.

## Changes
- `packages/opencode/src/index.ts` — `showQuotaToast` marks killswitch-blocked accounts as `blocked` and factors that into the active-account selection shown in the toast.

## Notes
Now that #35 and #36 are in `main`, this PR is just the toast-indicator delta on top of them.

### DIFF
--- a/packages/opencode/src/index.ts
+++ b/packages/opencode/src/index.ts
@@ -967,6 +967,10 @@ export const AnthropicAuthPlugin: Plugin = async (ctx) => {
       quota?: OAuthQuotaSnapshot
     }>,
     activeAccountId?: string,
+    isKilled?: (
+      quota: OAuthQuotaSnapshot | undefined,
+      accountId?: string,
+    ) => boolean,
   ) {
     const sections: string[] = []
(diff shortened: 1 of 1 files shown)


iceteaSA:
Updated: this branch now carries the sidebar quota fix (#57), merged in to keep the sidebar-touching branches consistent. #57's commits appear in this PR's diff until #57 merges to `main`, after which the diff reduces to this PR's own changes. **Merge order: #57 first.**

ualtinok:
Closing this because the quota toast and killswitch implementation has since been reworked, including model-scoped thresholds and block classification. This older indicator path no longer matches the current behavior. Thanks for the contribution.

PR #49, opened 2026-05-29
fix(opencode): repair thinking blocks rejected by Anthropic API

iceteaSA:
## Problem

Anthropic signs each thinking block against its original position. Two things break signatures and trigger `thinking blocks cannot be modified` 400s:

1. **Interleaved thinking flattened by the AI SDK** — opencode's Vercel AI SDK regroups all thinking blocks to the front of the assistant turn instead of leaving them interleaved with the tool_use blocks they authorized. The displaced blocks no longer match their signed positions.

2. **Stale signatures on historical assistant turns** replayed in a new request context (after context compaction or provider failover).

3. **Final block constraint** — Anthropic also rejects an assistant message whose final block is a `thinking` block.

## Strategy

- Keep a single signed thinking block only when it is the **lone block in the latest assistant turn** (the normal, still-valid current-turn case).
- Downgrade all others to `<thinking>` text — preserves reasoning content without invalid signatures.
- Drop `redacted_thinking` blocks (no readable text to preserve).
- Guard against a trailing `thinking` block being the final block in the message.
- Sanitize lone UTF-16 surrogates in downgraded text to avoid invalid-UTF8 400s.

## Cache impact

None on the static prefix. The transform runs before cache_control anchors are applied, so anchors land on the already-normalized structure. Historical turns render deterministically once they become non-last. The only cost is the unavoidable one-time tail transition when the current turn becomes historical.

## Tests

6 new test cases covering: flattened interleaving, single-block preservation, historical downgrades, redacted_thinking drops, lone-surrogate sanitization, and trailing-thinking-block downgrade.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/anthropic-auth/pr/49"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782672001&installation_id=135454708&pr_number=49&repository=cortexkit%2Fanthropic-auth&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fanthropic-auth%2Fpull%2F49&signature=daaee6bee201954b4da450d72e0d9c990bbb5d477c2c631387a82b501ec4086d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

---
## Summary by cubic
Repairs Anthropic 400s by normalizing `thinking` blocks in `opencode` requests. Prevents errors from flattened interleaving, stale signatures on historical turns, and final `thinking` blocks.

- **Bug Fixes**
  - Keep a single signed `thinking` block only in the latest assistant turn; downgrade all others to `<thinking>` text.
  - Drop `redacted_thinking` blocks unless it’s the single block in the latest assistant turn; never allow it as the final block.
  - Prevent `thinking` as the final block in an assistant message.
  - Sanitize lone UTF-16 surrogates in downgraded text.

<sup>Written for commit b6971c40864b953342a3c4e694fe15be8be0f081. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/anthropic-auth/pull/49?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/opencode/src/tests/transform.test.ts
+++ b/packages/opencode/src/tests/transform.test.ts
@@ -1306,3 +1306,190 @@ describe('sanitizeSystemText – realistic prompt', () => {
     expect(parsed).toMatchSnapshot()
   })
 })
+
+describe('rewriteRequestBody — thinking block repair', () => {
+  const userMsg = { role: 'user', content: 'hello world test message' }
+  const toolResult = {
+    role: 'user',
+    content: [{ type: 'tool_result', tool_use_id: 't1', content: 'out' }],
+  }
+
+  function assistant(content: unknown[]) {
+    return { role: 'assistant', content }
+  }
+
+  test('downgrades flattened interleaved thinking blocks to text', async () => {
+    const body = JSON.stringify({
+      messages: [
+        userMsg,
+        assistant([
+          { type: 'thinking', thinking: 'A', signature: 'sigA' },
+          { type: 'text', text: 'plan' },
+          { type: 'thinking', thinking: 'B', signature: 'sigB' },
+          { type: 'thinking', thinking: 'C', signature: 'sigC' },
+          { type: 'tool_use', id: 't1', name: 'bash', input: {} },
+        ]),
+        toolResult,
+      ],
+    })
+    const result = JSON.parse(await rewriteRequestBody(body))
+    const content = result.messages[1].content
+    // All thinking blocks downgraded; no signed thinking blocks remain.
+    expect(content.some((b: any) => b.type === 'thinking')).toBe(false)
+    expect(content[0]).toEqual({ type: 'text', text: '<thinking>A</thinking>' })
+    expect(content[1]).toEqual({ type: 'text', text: 'plan' })
+    expect(content[2]).toEqual({ type: 'text', text: '<thinking>B</thinking>' })
+    expect(content[3]).toEqual({ type: 'text', text: '<thinking>C</thinking>' })
+    // tool_use preserved (and prefixed).
+    expect(content[4].type).toBe('tool_use')
+    expect(content[4].name).toBe('mcp_Bash')
+  })
+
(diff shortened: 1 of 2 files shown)


ualtinok:
Closing this in favor of the upstream OpenCode fix: https://github.com/anomalyco/opencode/pull/30182.

This PR repairs the final Anthropic payload by downgrading/dropping `thinking` / `redacted_thinking` blocks. That can avoid 400s, but it is a broad and destructive sanitizer: it can discard valid signed/redacted thinking that should be preserved when the upstream message shape is correct.

The upstream PR addresses the signed-thinking reorder issue at the source by keeping OpenCode's existing ordinary-text reorder while skipping it when it would move signed or redacted Anthropic reasoning.

After that lands, we should update the OpenCode dependency and re-test. If there is still a concrete residual invalid-body case, we can open a narrower plugin-side guard for that specific shape.

PR #73, opened 2026-06-11
feat(opencode): shared tui-preferences.jsonc with customizable sidebar and forceToTop

iceteaSA:
## Summary

Adds a shared `tui-preferences.jsonc` for opencode TUI plugins, and makes the quota sidebar fully customizable through it.

- **New module** `packages/opencode/src/tui-preferences.ts` (subpath export `./tui-prefs`):
  - File resolution: `OPENCODE_TUI_PREFERENCES_FILE` override → `$OPENCODE_CONFIG_DIR` → `$XDG_CONFIG_HOME/opencode` → `~/.config/opencode`
  - Tolerant JSONC reads (malformed file → defaults, never crashes the TUI)
  - Per-key validation: numbers clamped, strings truncated, wrong types fall back individually
  - Atomic comment-preserving writes (jsonc-parser `modify`/`applyEdits`, temp + rename, serialized)
  - Debounced rename-safe directory watcher for live reload
  - `computeEffectiveOrder`: shared `forceToTop` convention — forced plugins float above everything (`-100000 + key index in file`), user `order` clamps to ±10000 so it can never beat a forced plugin
- **Sidebar** (`tui.tsx`):
  - Collapse state persists across restarts (`rememberCollapsed`, default on)
  - Prefs + collapse state live in a plugin-closure controller, surviving slot remounts (view switches)
  - Knobs: `startCollapsed`, `pollMs`, `refreshDebounceMs`, header label/version, per-section visibility (quota, fallbackAccounts, routing, cache, health), bar width/chars, warn/error thresholds
  - LIMITED badge always renders when degraded, regardless of preferences
  - Slot order from `forceToTop`/`order` at registration
- **Docs**: README section with the schema, limits, and the shared-file convention for other plugins
- **Tests**: 33 new unit tests (resolution, clamping, comment preservation, atomicity, watcher, ordering)

## Notes

- File is shared by convention: one top-level key per plugin, short non-integer-like name
- Layout knobs (`forceToTop`, `order`) apply at TUI startup; everything else hot-reloads
- Linux `fs.watch` reports the temp filename on rename — the watcher filter accounts for it

## Testing

- `bun test src/tests/tui-preferences.test.ts`: 33 pass
- Full suite: 434 pass / 1 pre-existing failure (provider.models cost test, unrelated)
- `typecheck`, `lint`, `build`, `pack --dry-run` all clean
- Manual: collapse → restart persists; live edits apply <1s; `forceToTop` pins above built-ins; subagent-view round-trip keeps collapse

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/anthropic-auth/pr/73"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783812827&installation_id=135454708&pr_number=73&repository=cortexkit%2Fanthropic-auth&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fanthropic-auth%2Fpull%2F73&signature=7acb594bfe0c94247dd22f6820072eabf82e959e92e2a3bf91ddf9b02b2e0651"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>

---
## Summary by cubic
Added a shared `tui-preferences.jsonc` and made the OpenCode Anthropic TUI sidebar customizable with persisted collapse and `forceToTop` ordering. Preferences are read safely, update live, and never crash the TUI.

- New Features
  - Shared preferences file with tolerant JSONC reads, comment-preserving atomic writes using `jsonc-parser`, debounced directory watcher for live reload, and config resolution (env override → XDG paths).
  - Helper APIs exported at `@cortexkit/opencode-anthropic-auth/tui-prefs`: `readTuiPreferencesFile`, `queueTuiPreferenceUpdate`, and `computeEffectiveOrder` implementing the `forceToTop` convention.
  - Sidebar is configurable: header label/version, start/persist collapse, section visibility, poll/debounce timing, quota bar width/chars, and warn/error thresholds; prefs and collapse state survive slot remounts.
  - README documents the schema and ordering convention; 33 unit tests added.

- Bug Fixes
  - Malformed or non-object preference files are ignored; per-key clamping/defaults prevent crashes.
  - Error threshold is enforced above warn; LIMITED badge always shows when degraded.
  - Watcher filters temp/sibling files, verifies content before firing, debounces bursts, and won’t revert a user’s collapse click (waits for our write to land).

<sup>Written for commit 81c3544cb263eba1e83f41dfadece0442ecb3045. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/anthropic-auth/pull/73?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,6 +316,84 @@ The sidebar polls plugin state and refreshes on OpenCode session and message eve
 
 Click the `CLAUDE` header to collapse or expand the sidebar. Collapsed, it shows the active account's 5-hour quota usage and a fast-mode row when fast mode is on; the header shows the plugin version (or a `LIMITED` badge when degraded). Collapse state is per-session and resets when OpenCode restarts.
 
+## TUI preferences
+
+The TUI sidebar reads `tui-preferences.jsonc` from the opencode config
+directory (`$OPENCODE_CONFIG_DIR`, else `$XDG_CONFIG_HOME/opencode`, else
+`~/.config/opencode`). The file is optional — defaults apply without it — and
+safe to hand-edit: the plugin updates single keys in place, preserving
+comments, and picks up edits live while the TUI runs.
+
+The file is shared: any opencode TUI plugin can claim one top-level key
+(short plugin name). This plugin uses `anthropic-auth`:
+
+```jsonc
+{
+  "anthropic-auth": {
+    // Layout — applied at TUI startup; restart to change
+    "forceToTop": false,        // jump to the first sidebar slot
+    "order": 160,               // slot order when not forced (-10000..10000)
+
+    // Behavior
+    "startCollapsed": false,    // initial state when nothing persisted
+    "rememberCollapsed": true,  // persist header clicks across restarts
+    "collapsed": false,         // written by the plugin on header click
+    "pollMs": 1500,             // state poll interval (500..30000)
+    "refreshDebounceMs": 200,   // event debounce (50..5000)
+
+    // Header
+    "header": {
+      "label": "CLAUDE",        // badge text (1..20 chars)
+      "showVersion": true
+    },
+
+    // Sections
+    "sections": {
+      "quota": true,
+      "fallbackAccounts": true,
+      "routing": true,
+      "cache": true,
+      "health": true
+    },
+
+    // Appearance
+    "appearance": {
+      "barWidth": 10,           // quota bar width (4..40)
+      "barFilledChar": "█",
+      "barEmptyChar": "░",
+      "warnThreshold": 50,      // used% where bars turn warn-colored
+      "errorThreshold": 80      // used% where bars turn error-colored
+    }
+  }
+}
+```
+
+Out-of-range numbers are clamped, wrong types fall back to defaults per key,
+and a malformed file is ignored entirely — the sidebar never crashes on bad
+preferences. The LIMITED badge always renders when an account is degraded,
+regardless of preferences.
+
+### forceToTop convention
+
+Sidebar slots render in ascending `order` (opencode built-ins use 100-500).
+Plugins adopting this convention compute their order as
+`-100000 + indexOfKeyInFile` when their `forceToTop` is `true`, so all forced
+plugins float above everything else, ordered by their key's position in the
+file. Reorder keys to reprioritize. Manual `order` values clamp to
+-10000..10000 and can never beat a forced plugin.
+
+Other plugins can reuse the helpers:
+
+```ts
+import {
+  computeEffectiveOrder,
+  readTuiPreferencesFile,
+} from '@cortexkit/opencode-anthropic-auth/tui-prefs'
+
+const root = await readTuiPreferencesFile()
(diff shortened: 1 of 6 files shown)


ualtinok:
Thanks for this. I tested it against current `main` (including the direct-dump and quota-race commits). It has a real watcher bug before I would merge it.

What I ran:
- temp merge onto current `main`; only conflict was the `solid-js` import in `packages/opencode/src/tui.tsx`
- resolved that import conflict and ran `bun run typecheck && bun test packages/opencode/src/tests/tui-preferences.test.ts && bun run build`

The focused preferences test fails:

```text
watchTuiPreferences > ignores sibling files that share the preferences name as a prefix
Expected: 0
Received: 1
```

I also verified the mechanism directly: on this platform, writing `tui-preferences.jsonc.backup` can emit an `fs.watch` event that looks like the real prefs file (`rename` for `tui-preferences.jsonc`) in addition to the backup filename. So filtering only by the reported filename is not sufficient.

Suggested fix: after the debounce fires, verify that the actual prefs file changed before calling `onChange` (for example stat mtime/ino/size, or a small content hash/cache of the target file). That should ignore sibling backup/temp files even when `fs.watch` coalesces or misattributes the filename.

iceteaSA:
Fixed and rebased onto current `main` (2519005).

**Watcher fix** (`81c3544`): you were right — filename filtering alone can't survive platform-dependent event misattribution. The watcher now uses a two-stage filter: the filename check stays as a cheap pre-filter, but after the debounce fires it re-reads the prefs file and compares content against the last-seen text; `onChange` only fires when the content actually changed. Sibling/backup writes that coalesce into a misattributed `rename` are dropped regardless of what `fs.watch` reports. Added a regression test that rewrites the file with byte-identical content (fresh event, same text) and asserts no callback — red on the old code, green now.

Focused suite: 34/34. typecheck/lint/build clean on the rebased branch.

PR #146, opened 2026-08-10
fix(opencode): keep client-side content-filter recovery armed in server fallback mode

iceteaSA:
Server-side safety fallback and the client-side content-filter recovery are currently mutually exclusive. `fallbackMode` defaults to `server`, and the client-side recovery is gated behind `fallbackMode === 'legacy'`, so in the default configuration a refusal that Anthropic's server-side fallback does not absorb reaches the client as a fatal `ContentFilterError` with no downgrade path — and the session stays broken for every subsequent turn.

Hit live on a Fable 5 session today:

```
22:10:50  assistant  ContentFilterError "The response was blocked by the provider's content filter"
22:13:29  INFO [fable-fallback] Anthropic server-side safety fallback active … targetModel=claude-opus-4-8 handoff=true
22:17:15  assistant  ContentFilterError
```

No recovery line anywhere, and the 22:13 fallback never logged "ended". An earlier refusal in the same session (`active` → `ended` in 12s) shows the server-side path working when Anthropic does absorb the refusal — the failure is specifically the unabsorbed path. Both refusals were triggered by benign content (a documentation URL, and a malformed `curl` command).

## Change

Removes the `fallbackMode === 'legacy'` gate from the four client-side recovery sites in `index.ts` (warm-chain re-plan, downgraded-body rewrite, `onContentFilter`, `onComplete`), making server-side fallback the first line and the client-side downgrade a backstop rather than an alternative.

This does not double-recover: when server-side fallback absorbs a refusal, the stream rewriter turns it into a normal completion, so `stop_reason: "refusal"` never reaches the finish-state handler and `onContentFilter` does not fire. The backstop only engages on the path that currently dies.

`serverFallbackModel` now derives from `fablePlan?.effectiveModel ?? requestModel` (and moved below the warm-chain re-plan so the plan is final). Without this, a downgraded turn would carry a `claude-opus-4-8` body while still requesting server-side fallback for the original model.

The default mode and the `OPENCODE_ANTHROPIC_AUTH_FALLBACK_MODE` escape hatch are unchanged.

## Tests

Three tests added, all in server (default) mode:

- refusal with no server-side handoff activates the client-side downgrade — the wedge regression. Fails on `main` (both requests carry `claude-fable-5`), passes here.
- an absorbed server-side fallback does not activate the client-side downgrade — no double-recovery.
- a downgraded Opus request does not carry the server-side fallback opt-in.

## Verification

typecheck, build, lint, biome check clean. opencode 1014 / core 69 / pi 57 pass.

e2e is 25 pass / 1 fail — `tool-prefix.test.ts:301` "bridges back to a stale Opus cache after more than 20 Fable blocks", which is **pre-existing on `main`**: verified 0/3 pass on a pristine `41e9aee` checkout with no commits from this branch, and 0/3 with them. Unrelated to this change.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/anthropic-auth/pr/146"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788986239&installation_model_id=22398&pr_number=146&repository=cortexkit%2Fanthropic-auth&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fanthropic-auth%2Fpull%2F146&signature=33bb899e8caa04152ce6d870a0046a27caf335021d836543e1d7f44166d1bd3e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Keeps client-side content-filter recovery armed in server fallback mode so unabsorbed refusals downgrade instead of throwing a fatal `ContentFilterError`. Also prevents cache prewarm requests from opting into server-side fallback routing in `opencode`.

- **Bug Fixes**
  - Removed the `fallbackMode === 'legacy'` gate for warm-chain replan, downgraded-body rewrite, `onContentFilter`, and `onComplete`.
  - Derived `serverFallbackModel` from `fablePlan?.effectiveModel ?? requestModel` after re-plan, so downgraded `claude-opus-4-8` requests don’t opt into server-side fallback.
  - Prevented double-recovery: absorbed server fallbacks are rewritten to normal completions, so client-side handlers do not fire.
  - Defaults unchanged: `fallbackMode` remains `server`; `OPENCODE_ANTHROPIC_AUTH_FALLBACK_MODE` still works.
  - Stripped `fallbacks` opt-in from source-model cache prewarm so warm requests reach the source model and don’t trigger server-side fallback routing.

<sup>Written for commit e8ebeaeb3750bfdd69312b70572840a3bc33c0a2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/anthropic-auth/pull/146?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

The PR keeps client-side content-filter recovery active as a backstop in server fallback mode while ensuring downgraded and prewarm requests use routing options appropriate to their effective models.
- Removes legacy-mode gating from fallback planning and recovery callbacks.
- Computes server fallback handling from the final effective request model.
- Strips server fallback opt-in from source-model cache prewarms.
- Adds regression coverage for unabsorbed refusals, absorbed server fallbacks, and downgraded requests.
</details>

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/opencode/src/index.ts | Enables client-side refusal recovery in server mode and aligns server fallback stream handling with the final planned request model. |
| packages/opencode/src/transform.ts | Removes the server-side fallback body opt-in when constructing source-model cache prewarm requests. |
| packages/opencode/src/tests/index.test.ts | Adds server-mode regression tests covering unabsorbed refusals, absorbed fallbacks, and downgraded request routing. |
| packages/opencode/src/tests/transform.test.ts | Verifies that cache prewarm transformation removes fallback and fast-mode opt-ins. |

</details>

<details open><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant Client
    participant Plugin
    participant Anthropic
    Client->>Plugin: Fable request
    Plugin->>Anthropic: Fable request with server fallback enabled
    alt Server fallback absorbs refusal
        Anthropic-->>Plugin: Normal rewritten completion
        Plugin-->>Client: Successful response
    else Refusal remains visible
        Anthropic-->>Plugin: refusal
        Plugin->>Plugin: Activate client-side downgrade
        Plugin-->>Client: ContentFilterError for current stream
        Client->>Plugin: Next request
        Plugin->>Anthropic: Opus request without server fallback opt-in
        Anthropic-->>Plugin: Completion
        Plugin-->>Client: Successful response
    end
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(opencode): strip server-side fallbac..."](https://github.com/cortexkit/anthropic-auth/commit/e8ebeaeb3750bfdd69312b70572840a3bc33c0a2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51960820)</sub>

### DIFF
--- a/packages/opencode/src/index.ts
+++ b/packages/opencode/src/index.ts
@@ -4143,27 +4143,25 @@ export const AnthropicAuthPlugin: Plugin = async (ctx) => {
                 incomingHeaders.get('x-session-affinity') ||
                 incomingHeaders.get('x-opencode-session')
               const requestModel = parseRequestModel(init?.body)
-              const serverFallbackModel =
-                fallbackMode === 'server' &&
-                isRecoverableRefusalModel(requestModel)
-                  ? requestModel
-                  : undefined
               let fablePlan = fableFallbackManager.plan(sessionId, init?.body)
-              if (
-                fallbackMode === 'legacy' &&
-                fablePlan &&
-                !fablePlan.downgraded
-              ) {
+              if (fablePlan && !fablePlan.downgraded) {
                 const finalWarm = recoveryWarmChains.get(fablePlan.recoveryKey)
                 if (finalWarm) {
                   await finalWarm
                   fablePlan = fableFallbackManager.plan(sessionId, init?.body)
                 }
               }
+              const serverFallbackModel =
+                fallbackMode === 'server' &&
+                isRecoverableRefusalModel(
+                  fablePlan?.effectiveModel ?? requestModel,
+                )
+                  ? (fablePlan?.effectiveModel ?? requestModel)
+                  : undefined
               const fableRequest: FableRequestContext | undefined = fablePlan
                 ? { plan: fablePlan }
                 : undefined
-              if (fallbackMode === 'legacy' && fablePlan?.downgraded) {
+              if (fablePlan?.downgraded) {
                 init = { ...init, body: fablePlan.bodyText }
               }
 
@@ -4178,9 +4176,7 @@ export const AnthropicAuthPlugin: Plugin = async (ctx) => {
                 createStrippedStream(response, {
                   perf: (stage, data) => trace.mark(stage, data),
                   contentFilterModel: fablePlan?.requestedModel,
-                  ...(fallbackMode === 'legacy' &&
-                  !fablePlan?.downgraded &&
-                  fablePlan
+                  ...(!fablePlan?.downgraded && fablePlan
                     ? {
                         onContentFilter: () => {
                           if (!fableRequest?.warmTarget) {
@@ -4218,9 +4214,7 @@ export const AnthropicAuthPlugin: Plugin = async (ctx) => {
                         },
                       }
                     : {}),
-                  ...(fallbackMode === 'legacy' &&
-                  fablePlan?.downgraded &&
-                  fableRequest
+                  ...(fablePlan?.downgraded && fableRequest
                     ? {
                         onComplete: (finishReason: string) => {
                           const completed = fableFallbackManager.complete(
--- a/packages/opencode/src/tests/index.test.ts
+++ b/packages/opencode/src/tests/index.test.ts
@@ -7917,6 +7917,278 @@ describe('auth.loader', () => {
     ).toBe(0)
   })
 
+  test('server mode — refusal with no server-side handoff downgrades to Opus (wedge regression)', async () => {
+    delete process.env.OPENCODE_ANTHROPIC_AUTH_FALLBACK_MODE
+    await useTempAccountFile(
+      createFallbackStorage({
+        accounts: [],
+        claudeCache: { enabled: false },
+        cacheKeep: { enabled: false },
+      }),
+    )
+    const models: string[] = []
+    let firstFable = true
+    const refusalSse = [
+      'event: message_start\ndata: {"type":"message_start","message":{"id":"msg_filtered"}}\n\n',
+      'event: message_delta\ndata: {"type":"message_delta","delta":{"stop_reason":"refusal"},"usage":{"output_tokens":0}}\n\n',
+      'event: message_stop\ndata: {"type":"message_stop"}\n\n',
+    ].join('')
+    const successSse = [
+      'event: message_start\ndata: {"type":"message_start","message":{"id":"msg_ok"}}\n\n',
+      'event: message_delta\ndata: {"type":"message_delta","delta":{"stop_reason":"end_turn"},"usage":{"output_tokens":1}}\n\n',
+      'event: message_stop\ndata: {"type":"message_stop"}\n\n',
+    ].join('')
+
+    globalThis.fetch = mock((input: any, init: any) => {
+      const url = extractUrl(input)
+      if (url.includes('/api/oauth/usage')) {
+        return Promise.resolve(
+          new Response(
+            JSON.stringify({
+              five_hour: { utilization: 0 },
+              seven_day: { utilization: 0 },
+            }),
+            { status: 200 },
+          ),
+        )
+      }
+      if (!url.includes('/v1/messages')) {
+        return Promise.resolve(new Response('{}', { status: 200 }))
+      }
+      const body = JSON.parse(String(init?.body)) as Record<string, unknown>
+      if (body.max_tokens === 0) {
+        return Promise.resolve(new Response('{}', { status: 200 }))
+      }
+      models.push(String(body.model))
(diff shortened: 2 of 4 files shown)


iceteaSA:
A reviewer raised three objections. One was correct and is now fixed; the other two I can refute with the request dumps and the session DB. Detail below so it's checkable rather than asserted.

## Fixed: the prewarm inherited the fallback opt-in

Correct, and it was a real defect introduced by this change. `prepareFableCacheWarmSource` rewrote `model` and deleted `speed` but left `fallbacks: 'default'` in the captured body, and `selectClaudeCodeBetas` re-adds `server-side-fallback-2026-07-01` whenever that key is `'default'` — so the source-model prewarm could itself be fallback-routed to Opus 4.8 and never warm the source cache. Harmless before this PR (the prewarm never ran in server mode); live once the recovery is armed there.

`e8ebeae` deletes `body.fallbacks` in the prewarm alongside `body.speed`, with a test asserting both the absent key and that `selectClaudeCodeBetas` no longer emits the beta. Red before, green after.

## The incident is real: 5 ContentFilterError turns, all on the source model

The claim that the session shows zero content-filter finishes doesn't match the database. Assistant messages for `ses_021c4e03…`, UTC:

```
20:10:44  claude-fable-5  ok
20:10:50  claude-fable-5  ContentFilterError
20:13:05  claude-fable-5  ok
20:13:29  claude-fable-5  ok          ← "server-side safety fallback active" logged here
20:14:31  claude-fable-5  ok
20:16:51  claude-fable-5  ok
20:17:15  claude-fable-5  ContentFilterError
20:19:37  claude-fable-5  ContentFilterError
20:20:16  claude-fable-5  ContentFilterError
20:21:55  claude-fable-5  ContentFilterError
```

Five `ContentFilterError` assistant messages, four of them after the last `active` event, each carrying `"The response was blocked by the provider's content filter"`. (Searching opencode's own log for the string finds only unrelated hits — my own tool-call arguments containing the identifier. The errors are persisted on the message rows, not in that log.)

## The refusals are on Fable 5, not on the fallback target

The objection that the test doesn't reproduce the incident rests on Opus 4.8 having been the active model that then refused. The request dumps say otherwise — every wire request in that session across the window:

```
28 × claude-fable-5
69 × claude-opus-5
 0 × claude-opus-4-8
```

Opus 4.8 was never sent. So the refusals are source-model refusals arriving with the fallback opt-in accepted but not acted on — which is exactly what the test exercises, and it is not a retry of a model that just refused. Each of those requests carried `fallbacks: "default"` and the `server-side-fallback-2026-07-01` beta and returned HTTP 200; there were 6 signed fallback markers in the session, so the mechanism was live throughout, not misconfigured.

The clearest evidence that the server-side path is best-effort is within one session, 7 minutes apart, identical request shape: `20:06:27 active → 20:06:39 ended` (absorbed, 12s) versus `20:13:29 active → never ended`, followed by four refusals surfacing to the client.

## On the architecture point

Fair that this changes documented rollback modes into stacked policies, and I should have said so explicitly. The reasoning: server-side fallback is best-effort by nature, so treating it as a *replacement* for client-side recovery leaves the unabsorbed path with no handler at all. The stacking is narrow — when the server-side path absorbs a refusal the rewriter converts it to a normal completion, so `stop_reason: "refusal"` never reaches the finish-state handler and the client-side hook cannot fire (there is a test for that). Happy to add a docs update to this PR describing the layering, or to split the doc change out, whichever you prefer.

Branch now at `e8ebeae`: opencode 1015 / core 69 / pi 57 pass, typecheck, build, lint, biome clean.

Issue #151, opened 2026-08-18
[BUG] cachekeep ticks can overlap and issue duplicate paid prewarms for the same target


iceteaSA:
`CacheKeepManager.start()` schedules `tick()` on a 60s interval with no in-flight guard:

```ts
this.timer = (this.options.setIntervalImpl ?? globalThis.setInterval)(
  () => {
    void this.tick().catch((error) => { ... })
  },
  CACHE_KEEP_TICK_MS,   // 60_000
)
```

`tick()` is async and awaits each due target **sequentially**:

```ts
for (const target of this.targets.values()) {
  if (target.cacheExpiresAt > dueAt) continue
  await this.prewarm(target, now)
}
```

Each prewarm can take up to `CACHE_KEEP_PREWARM_TIMEOUT_MS` (30s), and up to `CACHE_KEEP_MAX_TARGETS` (32) targets can be tracked. So a single tick is bounded by roughly 32 × 30s, not by 30s — two slow targets alone are enough to exceed the 60s interval and start a second overlapping tick.

When that happens, both ticks iterate the same map and both see the same target as due, because `cacheExpiresAt` is only advanced *after* the awaited request returns. The result is duplicate prewarm requests for one cachekeep cycle. Prewarms are real billed API calls, so this is duplicated spend, and it also doubles the load on an endpoint that is already slow — which is exactly the condition that triggered it.

The slow path is not hypothetical: `cachekeep tick failed {"error":"The operation timed out."}` appears in my plugin log, so 30s prewarm timeouts do occur in production. Two of those in one tick is sufficient.

**Fix direction:** guard `tick()` against re-entry with an in-flight flag or promise cleared in a `finally`, or replace the interval with a self-scheduling loop that only re-arms after the previous pass settles. Either way `stop()` should ensure a stale in-flight callback cannot mutate state afterwards.

Related but independent: #150 (a thrown prewarm aborting the tick) and #149 (failing prewarms retried forever at a fixed cadence). Same file, three separate defects; I've kept them apart so they can be triaged and fixed independently.

ualtinok:
Fixed on `main` in 4d30588e. Timer and manual CacheKeep ticks now share one in-flight promise, so overlapping ticks cannot issue duplicate prewarms or duplicate registry publications. Added a deferred-fetch regression test proving two concurrent ticks produce one request.

Issue #149, opened 2026-08-18
[BUG] cachekeep retries a failing prewarm at a fixed 5-minute cadence forever, with no backoff


iceteaSA:
Seen on a live install. A tracked cachekeep target whose prewarm keeps failing is retried every 5 minutes indefinitely — no backoff, no failure budget, and no stop once the real cache TTL has elapsed.

From my plugin log this evening, one session:

```
16:59:53 WARN [cachekeep] prewarm failed {"session":"ses_018f...","status":529}
17:04:50 WARN [cachekeep] prewarm failed {"session":"ses_018f...","status":529}
17:09:50 WARN [cachekeep] prewarm failed {"session":"ses_018f...","status":529}
... 14 consecutive failures, exactly 5 minutes apart ...
18:04:52 WARN [cachekeep] prewarm failed {"session":"ses_018f...","status":529}
```

14 retries spanning 65 minutes, across a 60-minute cache TTL. 45 such failures across 6 distinct sessions in the same log.

**Mechanism** — `packages/core/src/cachekeep.ts`

The failure branch reschedules by a constant:

```ts
target.cacheExpiresAt = now + CACHE_KEEP_PREWARM_LEAD_MS + 5 * 60_000
```

and `tick()` treats a target as due when `target.cacheExpiresAt <= now + CACHE_KEEP_PREWARM_LEAD_MS`. The lead term cancels, so the retry interval is a fixed 5 minutes regardless of how many times it has already failed. `pruneTargets()` only drops a target more than `2 * CACHE_KEEP_TTL_MS` past its expiry, and each failed attempt pushes expiry forward, so the target keeps itself alive.

Two consequences:

1. **No backoff against an overloaded upstream.** 529 means overloaded. Retrying on a fixed cadence, from every process running the plugin, is the opposite of what that status asks for.
2. **Retries continue past the point of usefulness.** Once more than `CACHE_KEEP_TTL_MS` has passed since the last success, the cache entry is gone. A later retry that succeeds is no longer refreshing a warm entry — it pays a full cache write. In the log above the last 5+ retries were past that line.

**Fix direction:** track consecutive failures per target and apply bounded exponential backoff with jitter, and stop retrying (or drop the target) once the elapsed time since the last success exceeds the real cache TTL, since there is no longer a warm entry to keep alive.

Happy to send a PR for this if you want it — I have one open for a separate `tick()` bug in the same file and can keep them independent.

ualtinok:
[...]

PR #155, opened 2026-08-18
fix(core): contain thrown prewarm errors in the cachekeep tick

### LINKED ISSUE #150: [BUG] A thrown prewarm aborts the whole cachekeep tick and blocks every later target permanently
`sendPrewarm()` issues the prewarm with `signal: AbortSignal.timeout(...)`. On a timeout or network error `fetch` **throws** rather than returning a result, and nothing between that throw and `start()`'s `.catch()` handles it:

- `sendPrewarm()` — `await fetchImpl(...)`, no try/catch
- `prewarm()` — `await this.sendPrewarm(target)`, no try/catch
- `tick()` — `await this.prewarm(target, now)` inside `for (const target of this.targets.values())`, no try/catch
- `start()` — `.catch()` logs `cachekeep tick failed`

So one throwing target aborts the entire tick. Reproduced with an injected `fetchImpl` that throws for target A only, with A, B, C all tracked and due:

```
tick THREW: TimeoutError: The operation timed out.
prewarm attempts: ["A"]        <- B and C never attempted
tick +1m attempts: ["A"]
tick +2m attempts: ["A"]
tick +3m attempts: ["A"]
```

It never recovers. The failing target's backoff assignment lives in the `!result.ok` branch, which a throw skips, so `cacheExpiresAt` is never advanced — A stays first-due and re-aborts the tick on every subsequent tick. B and C are starved until they age out as stale roughly two hours later, at which point cachekeep is silently doing nothing for them. `publishTrackedSessions()` at the end of `tick()` is also skipped, so the cross-process lease record stops being refreshed.

This is live, not theoretical — `TimeoutError: The operation timed out.` is what `AbortSignal.timeout` produces under Bun 1.3.14, and that exact string appears as `cachekeep tick failed` in my plugin log.

Worth noting the other two callers of this same throwing path both defend against it, which is what makes the tick loop look like an oversight rather than a decision:

- `warmFableAfterOpus` wraps `prewarmNow()` in try/catch — `packages/opencode/src/index.ts:1197-1225`
- `PrimeManager.fire` wraps its send in try/catch

**One trap for whoever fixes it:** `prewarm()` currently uses `result.status == null` to mean "unbuildable body, delete the tracked session". Mapping a thrown fetch onto that shape would make a transient network blip permanently drop a live session — worse than the current bug. The thrown case needs its own discriminator so it takes the backoff branch, not the delete branch.

I have a fix with tests ready and will open a PR shortly.

iceteaSA:
Fixes #150.

`sendPrewarm()` calls `fetch` with `signal: AbortSignal.timeout(...)`, which **throws** on timeout or a network error instead of returning a result. Nothing between that throw and `start()`'s `.catch()` handled it, so a single throwing target aborted the entire `tick()`:

```
prewarm attempts: ["A"]        <- B and C never attempted
tick +1m attempts: ["A"]
tick +2m attempts: ["A"]
tick +3m attempts: ["A"]
```

And it never recovered: the backoff assignment lives in the `!result.ok` branch that a throw skips, so the failing target's `cacheExpiresAt` was never advanced and it stayed first-due, re-aborting every subsequent tick. Later targets were starved until they aged out as stale ~2h later, and `publishTrackedSessions()` never ran, so the cross-process lease stopped refreshing.

## The change

- `sendPrewarm()` catches the fetch throw and returns `{ ok: false, reason, transient: true }`.
- `prewarm()` routes transient failures to the existing backoff branch instead of the delete branch.
- `tick()` wraps each `prewarm()` call defensively, applying the same backoff, so a future unanticipated throw cannot reintroduce head-of-line blocking.

The delete path was the trap here. `prewarm()` used `result.status == null` to mean "unbuildable body, drop the tracked session", so mapping a thrown fetch onto that shape would have made a transient network blip permanently delete a live session — worse than the bug being fixed. Hence the separate `transient` discriminator: `transient: true` is set at exactly one site, and the delete predicate is now `result.status == null && !result.transient`. Unbuildable bodies still delete; HTTP-status failures still back off, unchanged.

## Two things worth flagging

- **`prewarmNow()` is a behavior change for consumers.** It previously rejected on a fetch throw; it now resolves `{ ok: false, transient: true }`. The only in-repo consumer (`warmFableAfterOpus`) already handles both, but this is a public surface of `@cortexkit/anthropic-auth-core`, so it's a contract change worth knowing about.
- **`loadStorage()` at the top of `tick()` is still unguarded** and can abort a tick. I left it alone: it's pre-existing, `start()`'s `.catch()` handles it, it retries in 60s, and skipping a pass when the schedule can't be read is arguably right. Happy to wrap it if you'd rather.

## Verification

New tests assert that a throwing target does not prevent later targets from being prewarmed, that the throwing target is retained rather than deleted, that it backs off instead of retrying on the next tick, and that an unbuildable body still deletes its target.

Proven red before green — reverting the source change makes the new tests fail (2 fail, with the unbuildable-delete test correctly still passing as a regression guard). Gates on the branch: `bun run typecheck` clean, `bun run test` 1025 pass / 0 fail, `bun run lint` clean.

Two other defects in this file are filed separately — #149 (failing prewarms retried at a fixed cadence forever) and #151 (ticks overlapping into duplicate paid prewarms). Kept independent so they can be triaged on their own; happy to send PRs for either if you want them.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/anthropic-auth/pr/155"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789677083&installation_model_id=22398&pr_number=155&repository=cortexkit%2Fanthropic-auth&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fanthropic-auth%2Fpull%2F155&signature=bb1b114086ccd43af36e14d19801274206b350d3df52079e05de7e17bc7dc163"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Contain thrown prewarm errors during cachekeep ticks to prevent head-of-line blocking and session starvation. Previously, a timeout/network error from `fetch` in `sendPrewarm()` threw and aborted the entire `tick()`; now these errors are caught, treated as transient failures, and the tick proceeds to later targets.

**Behavior changes and migration**
- `sendPrewarm()` catches `fetch` throws and returns `{ ok: false, reason, transient: true }`; `prewarm()` treats transient failures as backoff, not delete; `tick()` wraps per-target calls to apply the same backoff on unexpected throws.
- `prewarmNow()` no longer rejects on network/timeout errors. It resolves a failure result with `transient: true`. Migration: update callers to check `result.ok` (and optionally `result.transient`) instead of catching a rejection.
- Delete-on-prewarm now only applies to unbuildable bodies (`status == null` and not transient). Transient/network failures never delete tracked sessions.

<sup>Written for commit e9afa64043be6da6bc6c1790e64ff8e0786881a1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/anthropic-auth/pull/155?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

The PR contains fetch failures within cachekeep prewarming so one failed target no longer aborts the tick or starves later targets.
- Adds an explicit transient-failure result for fetch exceptions.
- Preserves transient targets and advances them using the existing retry backoff rather than deleting them.
- Defensively isolates each target during a tick and adds regression coverage for continuation, retention, backoff, publication, and unbuildable-body deletion.
</details>

The PR appears safe to merge, with the changed failure paths preserving target lifecycle semantics while preventing one prewarm failure from aborting the entire tick.

The implementation distinguishes transient fetch failures from unbuildable requests, applies retry backoff without deleting live targets, continues processing later targets, and retains the existing deletion behavior for invalid prewarm bodies.

| Filename | Overview |
|----------|----------|
| packages/core/src/cachekeep.ts | Converts thrown fetch failures into transient results, preserves affected targets with backoff, and isolates unexpected per-target failures so each tick can continue and publish state. |
| packages/opencode/src/tests/cachekeep.test.ts | Updates timeout expectations for the resolved transient result and adds coverage for target continuation, retention, backoff, publication, and unbuildable-body deletion. |

</details>

<details open><summary><h3>Flowchart</h3></summary>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
  T[Cachekeep tick] --> D{Target is due?}
  D -- No --> N[Check next target]
  D -- Yes --> P[Build and send prewarm]
  P --> S{Result}
  S -- Success --> U[Advance cache expiry]
  S -- HTTP or transient failure --> B[Apply retry backoff]
  S -- Unbuildable body --> X[Delete tracked target]
  P -- Unexpected throw --> C[Log and apply defensive backoff]
  U --> N
  B --> N
  X --> N
  C --> N
  N --> F[Publish tracked sessions]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(core): contain thrown prewarm errors..."](https://github.com/cortexkit/anthropic-auth/commit/e9afa64043be6da6bc6c1790e64ff8e0786881a1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54515618)</sub>

**Context used:**

- Knowledge Base — [Core Package: Provider Abstraction, Model Catalog, and Request Shaping](https://app.greptile.com/cortexkit/-/custom-context/knowledge-base/cortexkit/anthropic-auth/-/docs/core-index-provider.md)

### DIFF
--- a/packages/core/src/cachekeep.ts
+++ b/packages/core/src/cachekeep.ts
@@ -339,7 +339,7 @@ export type CacheKeepPrewarmResult =
         cache_read_input_tokens?: number
       }
     }
-  | { ok: false; reason: string; status?: number }
+  | { ok: false; reason: string; status?: number; transient?: true }
 
 export class CacheKeepManager {
   private readonly targets = new Map<string, CacheKeepTarget>()
@@ -576,7 +576,15 @@ export class CacheKeepManager {
     const dueAt = now + CACHE_KEEP_PREWARM_LEAD_MS
     for (const target of this.targets.values()) {
       if (target.cacheExpiresAt > dueAt) continue
-      await this.prewarm(target, now)
+      try {
+        await this.prewarm(target, now)
+      } catch (error) {
+        logger.warn('cachekeep', 'prewarm failed', {
+          session: target.id,
+          reason: error instanceof Error ? error.message : String(error),
+        })
+        target.cacheExpiresAt = now + CACHE_KEEP_PREWARM_LEAD_MS + 5 * 60_000
+      }
     }
     this.publishTrackedSessions()
   }
@@ -597,14 +605,23 @@ export class CacheKeepManager {
       : new Headers(target.headers)
     headers.delete('content-length')
     headers.delete('transfer-encoding')
-    const response = await fetchImpl(target.url, {
-      method: 'POST',
-      headers,
-      body: prewarm.bodyText,
-      signal: AbortSignal.timeout(
-        this.options.prewarmTimeoutMs ?? CACHE_KEEP_PREWARM_TIMEOUT_MS,
-      ),
-    })
+    let response: Response
+    try {
+      response = await fetchImpl(target.url, {
+        method: 'POST',
+        headers,
+        body: prewarm.bodyText,
+        signal: AbortSignal.timeout(
+          this.options.prewarmTimeoutMs ?? CACHE_KEEP_PREWARM_TIMEOUT_MS,
+        ),
+      })
+    } catch (error) {
+      return {
+        ok: false,
+        reason: error instanceof Error ? error.message : String(error),
+        transient: true,
+      }
+    }
     if (!response.ok) {
       return {
         ok: false,
@@ -632,7 +649,7 @@ export class CacheKeepManager {
   private async prewarm(target: CacheKeepTarget, now: number) {
     const result = await this.sendPrewarm(target)
     if (!result.ok) {
-      if (result.status == null) {
+      if (result.status == null && !result.transient) {
         logger.debug('cachekeep', 'prewarm skipped', {
           session: target.id,
           reason: result.reason,
--- a/packages/opencode/src/tests/cachekeep.test.ts
+++ b/packages/opencode/src/tests/cachekeep.test.ts
@@ -314,7 +314,7 @@ describe('CacheKeepManager', () => {
       prewarmTimeoutMs: 5,
     })
 
-    const result = manager.prewarmNow({
+    const result = await manager.prewarmNow({
       sessionId: 'ses_timeout',
       url: 'https://api.anthropic.com/v1/messages?beta=true',
       headers: new Headers(),
@@ -330,7 +330,10 @@ describe('CacheKeepManager', () => {
       }),
     })
 
-    await expect(result).rejects.toMatchObject({ name: 'TimeoutError' })
+    expect(result.ok).toBe(false)
+    if (result.ok) throw new Error('expected prewarm to fail')
+    expect(result.transient).toBe(true)
+    expect(result.reason).toContain('timed out')
   })
 
   test('passes OAuth account identity to prewarm header refresh', async () => {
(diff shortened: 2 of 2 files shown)


ualtinok:
Integrated onto current `main` as `fac7e29c`, preserving the original author. Verified after rebuilding core: 1,088 OpenCode tests and 63 Pi tests pass, along with workspace typecheck, lint, and formatting checks. Thank you.

PR #219, opened 2026-09-16
fix(pi): isolate the test suite from the host config directory

iceteaSA:
`packages/pi` has no test preload, so its suite inherits the host's config directories. One test reads the operator's live `~/.config/opencode/anthropic-auth.json` and passes because of what happens to be in it — on a machine without that file it fails.

That is CI, and anyone who clones the repo.

## Evidence

`packages/pi/src/tests/stream.test.ts` → `sticky-balanced preserves the strict API fallback gate after OAuth exhaustion`, varying one variable at a time on a developer machine:

```
baseline                            1 pass 0 fail
HOME=<empty tmpdir>                 0 pass 1 fail
XDG_CONFIG_HOME=<empty tmpdir>      0 pass 1 fail
OPENCODE_CONFIG_DIR=<empty tmpdir>  0 pass 1 fail
PI_AGENT_DIR=<empty tmpdir>         1 pass 0 fail
```

Pi's own agent dir is not the dependency. The OpenCode config dir is: `getConfigDir()` in `packages/core/src/accounts.ts` resolves `$OPENCODE_CONFIG_DIR`, else `$XDG_CONFIG_HOME/opencode`, else `~/.config/opencode`, and Pi tests reach core code that falls back to it.

Pre-existing, and not caused by any recent change — a clean detached worktree at `main` fails identically under an empty `HOME`. `packages/core` and `packages/opencode` each received this preload on 2026-09-07; `packages/pi` was missed.

## Change

`packages/pi/bunfig.toml` + `packages/pi/src/tests/setup.ts`, mirroring the existing two setups: a per-test `mkdtemp`, host path variables deleted and redirected into it, temp dir removed.

Seven variables, chosen by reading the call graph rather than by guessing — `OPENCODE_CONFIG_DIR`, `OPENCODE_ANTHROPIC_AUTH_FILE`, `OPENCODE_ANTHROPIC_AUTH_STATE_FILE` (`accounts.ts:514-528`), `PI_AGENT_DIR` (`paths.ts:7`), and the three `PI_ANTHROPIC_AUTH_*` state vars (`stream.ts:138,157`) which are reset rather than pinned since tests set them individually.

The preload deliberately does **not** restore the host's original values. The first version did, in `afterEach` — but Bun runs `afterEach` inner-file-first and the preload's hook last, so a per-file `afterAll` saw host values again, and `commands.test.ts:82` asserts `OPENCODE_ANTHROPIC_AUTH_STATE_FILE` is unset. Any host exporting that variable got a red suite from the isolation meant to make it deterministic:

```
with OPENCODE_ANTHROPIC_AUTH_STATE_FILE set   114 pass  1 fail
after the fix                                 114 pass  0 fail
```

Restoring is unnecessary anyway: this is in-process mutation that cannot reach the operator's shell, and every `beforeEach` re-isolates. Dropping the snapshot removed a second hazard with it — its entries were only cleared in `afterEach`, so a test that threw could leave a stale value to restore over a later test's isolation.

## Verification

```
before (main, empty env)   0 pass  1 fail
after  (empty env)       114 pass  0 fail
after  (normal env)      114 pass  0 fail
```

Empty env means `HOME`, `XDG_CONFIG_HOME`, `OPENCODE_CONFIG_DIR` and `PI_AGENT_DIR` all pointed at a fresh temp dir. The before/after pair is what shows the preload is load-bearing rather than incidental.

core 188/188 · opencode 1869/1869 · pi 114/114 · typecheck, format, biome clean.

No production file and no test assertion was modified.

## Note

`OPENCODE_CONFIG_DIR` alone was not sufficient in practice — the storage and state file variables had to be pinned too. The isolation is proven by the suite passing in a clean environment; the precise call site that still reached the host path was not pinned down by reading, and is recorded here rather than left implied.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/anthropic-auth/pr/219"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792150543&installation_model_id=22398&pr_number=219&repository=cortexkit%2Fanthropic-auth&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fanthropic-auth%2Fpull%2F219&signature=7a4bf057fe3647fa21320172fb0fae43876d501dd637a3d8d418a60cb1b2c69b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Isolates the `packages/pi` test suite from the host's config directories so tests pass in CI and on machines without a live OpenCode config.

- Adds a preload that redirects the OpenCode and Pi config path variables into a per-test temp dir without restoring host values afterward — the restore re-exposed host config between tests, and the env mutation is in-process only, so it can't reach the operator's shell.
- Mirrors the existing isolation in `packages/core` and `packages/opencode`; `packages/pi` was missed when they received theirs.

<sup>Written for commit 9f59e0514ed7754eb7ed7dd95656ac8a7e719a8b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/anthropic-auth/pull/219?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


iceteaSA:

```
```

This PR: base of the stack. Its own check is red on `bridges back to a stale Opus cache…`, which is a **pre-existing** failure on main, not something this PR introduces — it is fixed in #229. Merging this first does not regress anything.

Issue #220, opened 2026-09-16
Two async-timing tests are flaky on CI: identical trees give different failures


iceteaSA:
Two async-timing tests fail intermittently on CI. Same bytes in, different outcome each run — so a red `check` on a PR currently carries no information about the PR.

## Evidence

Three CI runs of the **same tree** (commits differ only by amend; `git diff` between them is empty), PR #218:

| run | commit | result |
|---|---|---|
| 1 | `b90645b` | FAIL — `packages/e2e-tests` › `bridges back to a stale Opus cache after more than 20 Fable blocks` |
| 2 | `4dfbf7f` | FAIL — `packages/opencode` › `writes the manifest before clearing a migrated legacy handle` |
| 3 | `e2af7f6` | PASS |

Two different tests, then green, from identical source.

## The two failures

**`bridges back to a stale Opus cache…`** — `waitForSessionText` gives up after ~36–42 s waiting for the `Fable recovery window complete. Returning to Fable 5.` notice (`packages/opencode/src/index.ts:783`). It passed on another CI run in 16 s, so the wait is racing notice delivery rather than observing a stuck one.

**`writes the manifest before clearing a migrated legacy handle`** (`packages/opencode/src/tests/index.test.ts:2451`) — asserts on `stateAtManifestWrite`, captured inside an `fs.rename` spy. It failed with `Received value must be an array type…`, i.e. the capture was still `undefined`: the startup migration had not reached its manifest write by the time `getPlugin` returned and the assertion ran.

Both are the same shape — an assertion racing detached async work, with no barrier tying the two together.

## Not reproducible locally

`writes the manifest…` passes 5/5 in isolation and inside full-suite runs (1874/1874, repeatedly). The e2e test fails consistently on my machine — but it fails identically on a clean detached worktree at `main`, so that is a local environment issue, not a signal about either PR.

## Suggested fix

Same treatment as #176: replace the implicit timing assumption with an explicit barrier. For the migration test, expose or await the startup migration's completion rather than relying on it finishing before `getPlugin` resolves. For the e2e test, wait on the notice being *queued* rather than on its rendered text, or raise the bound with a named timeout that says what it was waiting for.

Happy to take either if useful — filing first since both are pre-existing and neither is related to what I am working on.

iceteaSA:
Follow-up with a firmer diagnosis on the e2e half — it is **not** a timeout that needs raising.

[...]

PR #228, opened 2026-09-16
test(custody): make manifest-lock contention tests independent of wall clock

iceteaSA:
Six manifest-lock and legacy-migration tests fail under CPU contention and pass on a quiet machine. One of them (`reports a held lock as lock_busy`) now fails on CI and blocks unrelated PRs; filed as #220.

**The coupling.** In `withCustodyManifestLock`, `deadline = startedAt + ttlMs` is the *contender's* give-up deadline, while `now() - ownerClaimedAtMs >= ttlMs` is the *holder's* staleness threshold. One knob, two opposed roles. The test set `ttlMs: 30`, so a holder starved past one renewal tick looked stale and the contender evicted it — acquiring the lock instead of reporting busy, which is a semantic failure, not a slow test.

Raising the TTL does not fix it, and I tried: at 2s the contender still evicted under load, *and* the longer give-up path started tripping bun's 5s per-test timeout. Any fix expressed as a duration is the same defect one level down. These assertions are about ordering, not elapsed time.

**The fix** is a synthetic held owner plus a scripted injected clock, so "holder is fresh" and "contender exhausted its deadline" stop competing for one wall clock. The migration tests raced a different deadline — the 100ms startup warmup — and the concurrent one used 1s sleeps, now explicit barriers. Production code is untouched: the diff is two test files.

**The failure mode changed class**, which is the claim worth checking. Under 16 CPU hogs, six runs each:

```
before   4/2  3/3  0/6  1/5  4/2  1/5    assertion failures in every run (2,2,5,2,2,2)
after    6/0  5/1  6/0  6/0  5/1  6/0    assertion failures: 0 everywhere
```

After the change, every red is a watchdog kill and its ENOENT cleanup cascade — never an assertion. Quiet: 10/10 green, 1.04–30.42ms.

**What this does not fix.** Extreme oversubscription still trips bun's 5s watchdog on arbitrary tests, because the OS deschedules the test process. The evidence is a three-syscall `lock_busy` test reported at **7588ms** — that is scheduler starvation, not a timing dependence in the code. It is a harness property and it is why the tally above is 5/6 rather than 6/6 under deliberate load. Relevant to #220: CI runners are 2-core, so a red check there deserves this reading before it is blamed on a diff.

Gates: core 188/0 · opencode 1869/0 · pi 114/0 · typecheck, format, biome clean.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/anthropic-auth/pr/228"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792178090&installation_model_id=22398&pr_number=228&repository=cortexkit%2Fanthropic-auth&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fanthropic-auth%2Fpull%2F228&signature=a5cd0ddd5d10479c3e04843a31d23bc2b9f220a4d54a38d8af6b7c6c825a825e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Makes the manifest-lock and legacy-migration tests deterministic by removing wall-clock dependence, and isolates the `packages/pi` test suite from the host config directory. The old lock tests used the TTL as both the contender's give-up deadline and the holder's staleness threshold, so a starved holder was evicted instead of reporting `lock_busy` under CPU load (issue #220); the PI suite had no test preload and one test read the operator's live config, passing only when that file happened to exist.

- Replaces elapsed-time waits with a synthetic owner plus injected clock in the lock-busy test, and explicit barriers in the concurrent migration test.
- Adds a `packages/pi` test preload that points config env vars at a fresh temp dir per test; host values are deliberately not restored.
- Production code is untouched; only test files and `packages/pi/bunfig.toml` change.
- Does not fix bun's 5s watchdog kills under extreme oversubscription, so a CI red on a 2-core runner with ~7s reported times means scheduler starvation, not a semantic failure.

<sup>Written for commit 6a00dbb6a1de4606ec7aed2b065c84e622965aef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/anthropic-auth/pull/228?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


iceteaSA:
First CI run confirmed the fix on the real runner — **all six target tests passed**, including `reports a held lock as lock_busy`, which is the one that had been failing there.

The red was an unrelated pre-existing failure: `Pi API fallback routing helpers > sticky-balanced preserves the strict API fallback gate after OAuth exhaustion`. `packages/pi` has no test preload, so its suite reads the operator's live `~/.config/opencode/anthropic-auth.json` and passes on a machine that happens to have one. That is fixed in #219, so I have rebased this branch on top of it. Both this PR and #218 now stack on #219.

Zero assertion failures and zero watchdog kills in that run — the classification this change is built around held on CI, not just under my synthetic load.

core 188/0 · opencode 1869/0 · pi 114/0 · typecheck clean.

iceteaSA:

```
```

This PR: stacked on #219. Green. Fixes the manifest-lock flake that was blocking #218.

Issue #162, opened 2026-08-22
e2e: stale-Opus bridge test fails on main since 80fb2b7 (desktop notices from idle events)


iceteaSA:
`packages/e2e-tests` → `bridges back to a stale Opus cache after more than 20 Fable blocks` fails on current `main` (`5fed3b9`), and bisects to `80fb2b7` (`fix(opencode): deliver desktop notices from idle events`).

Bisect evidence (fresh worktree per ref, `bun install --frozen-lockfile` + `bun run build` before each run):

```
4d30588  fix(core): harden cachekeep and fallback state handling   → 1 pass / 0 fail
80fb2b7  fix(opencode): deliver desktop notices from idle events   → 0 pass / 1 fail
5fed3b9  current main                                              → 0 pass / 1 fail
```

The failing assertion is a `toContain` on the bridge continuation (`Run bun run test:e2e` fails identically in CI — PR #160's latest run hit it on a branch whose only delta from main is the lane-start commit, which touches none of the notice path).

The test exercises the legacy Fable→Opus recovery bridge, whose flow ends in a desktop recovery notice — consistent with the notice-delivery timing change in `80fb2b7` (idle-event delivery vs the prior arming path) shifting when the notice flushes relative to the bridge turn the test asserts on.

Happy to help narrow further if useful; flagging now mainly so #160's red `check` is attributable — its e2e failure reproduces byte-identically on pure `main`.

iceteaSA:
Still reproduces on `a16087f` (current main, includes `6813762` "finalize notifications and response artifacts") — fresh worktree, frozen lockfile, full build: `0 pass / 1 fail` on the bridge test. So the idle-notice finalization pass didn't cover this path.

iceteaSA:
Dug into this further. Separating what's observed from what's inferred, since the last step is your design intent rather than something I can settle from the code.

**Observed** (e2e trace on `abc9972`, failing test, 3/3 deterministic):
- The assertion fails with expected `"second Opus recovery"` / received `"ok"` — i.e. the final prompt gets the mock's default response because the scripted queue has been consumed one entry early.
- There is an **extra non-zero-output Opus request that duplicates an earlier user prompt** (`continue Opus cycle 3`). It consumes a scripted Opus completion, so the second refusal lands during `healthy Fable turn 10` instead of the explicit final prompt.
- `6813762` defers the switched notice until retry completion but does not prevent that extra request.

**Inferred** (plausible, not proven — needs your read):
[...]

PR #229, opened 2026-09-16
fix(recovery): deliver the restoration notice after the final recovery turn

iceteaSA:
After a Fable/Opus recovery window closes, the `Fable recovery window complete. Returning to Fable 5.` notice never reaches a session with no TUI attached. The switch notice arrives; only the restoration one is lost. In production that means **the user is never told the model was restored** unless they happen to send another message — a human sends another turn and sees the state resolve anyway, which is why only the e2e ever noticed.

Not a slow test: at a 120s wait the notice still never arrived.

**Mechanism** (this refuted my first hypothesis, which was simply "no later idle event to release it"). The restoration notice *is* queued with a valid lease and its probe *does* run. The decisive part is overlap with the still-in-flight switch notice: OpenCode publishes our own ignored insertion as a user-message update plus a busy status. The lease logic could not tell that apart from a genuine user prompt, so notice A's delivery revoked notice B's lease — and with no further idle event coming, nothing ever released it.

**Fix.** Plugin-generated notice IDs are recorded and excluded from the genuine-user revision counter, so our own insertions no longer revoke the lease. Genuine user messages still bump the revision and still revoke it immediately.

**Why a duplicate billed provider turn remains impossible** — the hazard this path was reworked for in #162, where an ignored notice becoming the latest user message makes the run loop invoke the provider again on the same turn:
- re-entry after a successful insertion requires the genuine-user revision to be *unchanged*; any real user message blocks it,
- delivery still passes the existing live `session.status()` idle probe with its bounded re-arm,
- `canSend` still re-checks the lease immediately before insertion, so a user message landing mid-flight aborts the send.

Verified by mutation rather than by reading:
- forcing `isDesktopNotice = false` (pre-fix behaviour) reddens the e2e with `session notification not found: Returning to Fable 5` — the fix is load-bearing,
- removing the `canSend` fence reddens `downgrades a filtered Fable session for ten successful Opus turns and warms Fable after each` — the duplicate-turn guard is covered by a test, not merely present.

The e2e went from failing deterministically on this host (~36s) to passing in 13.8s.

Also adds debug logging at queue and flush-decision. That path had none, which is why diagnosing it required instrumenting the harness by hand.

Stacked on #219, which fixes an unrelated pre-existing Pi failure this branch would otherwise hit.

core 188/0 · opencode 1869/0 · pi 114/0 · e2e 36/0 · typecheck, format, biome clean.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/anthropic-auth/pr/229"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792180243&installation_model_id=22398&pr_number=229&repository=cortexkit%2Fanthropic-auth&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fanthropic-auth%2Fpull%2F229&signature=071149b41437715c7cf03808cf67769b73969aa58796e9cce348a95c349df110"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes the `Fable recovery window complete. Returning to Fable 5.` notice being silently dropped for headless sessions after a recovery, and makes the manifest-lock contention tests deterministic under CPU load.

**Bug Fixes**
- The restoration notice was queued but never delivered because the in-flight switch notice's own user-message update revoked its lease; plugin-generated notice IDs are now tracked separately from genuine user messages.
- Genuine user messages still revoke the lease immediately, and delivery still requires an unchanged user revision, the live idle probe, and a final lease check before insertion, so an active prompt can't adopt the notice and duplicate a billed provider turn.
- Adds debug logging to the desktop-notice queue and flush-decision path.

**Tests**
- Manifest-lock contention tests now use injected clocks and explicit barriers instead of wall-clock timing and sleeps.

<sup>Written for commit e8c31f37acf38f0c6e5ed0c3142d0fd36a4bb805. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/anthropic-auth/pull/229?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


iceteaSA:

```
```

This PR: stacked on #219. Green. Fixes the restoration-notice bug that was failing the e2e in #219.

iceteaSA:
Re-stacked into a single linear chain so every PR can go green independently:

```
#219  fix(pi): isolate the test suite from the host config
  └── #228  test(custody): lock tests independent of wall clock
        └── #229  fix(recovery): deliver the restoration notice
              └── #218  fix(custody): carry the vault credential id
```

Merge in that order. The earlier shape had #228/#229/#218 as siblings on #219, which left #218 red on the e2e notice bug that #229 fixes — a failure that was never its own.

Local gates on the chain tip: core 199/0 · opencode 1875/0 · pi 114/0 · e2e 36/0 · typecheck clean.

ualtinok:
Closing as superseded by the narrower fix already on `main` in `37b391b1`. That fix cancels superseded in-flight notices, continues delivery with the newest notice, and distinguishes plugin-injected ignored messages from genuine prompts without adopting the broader state-machine changes here. The deterministic regression was mutation-tested; the E2E suite passed 20/20 on OpenCode 1.18.18, 10/10 on 1.18.30, and again in the final full 36-test run.

PR #234, opened 2026-09-17
ci: record test-count floors (data only, no gate)

iceteaSA:
First of two. This adds only the floors data file — no workflow change, so nothing new runs in CI.

## Why this is split from the gate

The gate PR that follows compares a branch's floors against **the merge target's** floors. On its own PR the merge base would have no floors file, so the ratchet would necessarily take its UNCHECKED arm and block the merge.

The obvious fix is a one-time bootstrap exemption. I declined it: "base resolves but floor file absent" is not actually a one-time condition — it also occurs when someone *deletes* the floors file, which is exactly the bypass the ratchet exists to prevent. An exemption would be a permanent code path and a permanent reviewer-trust surface to solve a condition that happens once.

Landing the data first means that condition never arises, and "base has no floor file" stays a pure UNCHECKED forever.

## The numbers

Measured on a clean tree at `f9c867d1`, `git status --porcelain` empty:

```
core 199 · opencode 1885 · pi 114
```

The file records the HEAD and dirty-path count at measurement time. A floor carrying an unexpected HEAD or a non-zero dirty count is self-refuting on inspection; a bare number is not.

That stamp is not decorative. An earlier measurement in this work returned `202` twice for core — two agreeing runs, which was the discipline I had asked for — and both were wrong, because the tree had been switched underneath the measurement. **Agreement between two runs proves they shared a subject, not that the subject was right.** The stamp would have shown it.

## Evidence classes

The file states, per package, how each floor was established: `core` is backed by two agreeing runs *and* an independent git-derived delta against another ref; `opencode` and `pi` rest on double measurement alone. A reader deciding whether to trust a floor deserves to know which kind of evidence backs each number, and where to look first when one of them is wrong.

The note also records why a derivation is not an oracle: it is independent of the *run* but not of the framework's counting convention, and this repo has 16 `test.each` sites where declaration-counting and runner-counting disagree.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/anthropic-auth/pr/234"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792248004&installation_model_id=22398&pr_number=234&repository=cortexkit%2Fanthropic-auth&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fanthropic-auth%2Fpull%2F234&signature=dacff050aabeb5e4fb97a84fbebf98cb7293291cd2b91aee9f209d067b7a012d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Adds `.ci/test-count-floors.json` with baseline test counts for `core` (199), `opencode` (1894), and `pi` (114) so the follow-up gate PR has floors to compare against on `main`. This PR is data only — no workflow changes, nothing new runs in CI.

- The file records the HEAD and dirty-path count at measurement time, so a contaminated measurement is self-refuting.
- Evidence per floor says how each number was established: `core` was double-measured plus an independent git-derived delta, `opencode` double-measured plus a per-file decomposition, and `pi` double-measured only.
- Landing the data first keeps the gate PR’s merge target from having no floors file, avoiding a bootstrap exemption that would also forgive deleting the floors file.

<sup>Written for commit 3d8a1d10a24ede0d06cf220b2477b250498c5bd7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/anthropic-auth/pull/234?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


iceteaSA:
@ualtinok, following up. Rebased onto current `main` (`4a01612`); it was 15 commits behind. It's still data only, and #235 is the gate that reads it.

**The recorded floors are out of date.** They were measured at `1b27511a`, before v1.23.0: core **199**, opencode **1894**, pi **114**. `main` now runs core **295** and opencode **1926**; pi is unchanged. They're safe as they stand, because the gate only fails if a count drops *below* its floor, but they no longer protect the tests v1.23.0 added.

Before I refresh them, one point to settle on #235: the counts need to come from a **green** run. The ratchet refuses to count a failing run (correctly), and on my machine `main`'s opencode suite has an in-suite failure that CI doesn't (#220). So I'd take the new floors from a green CI run on `main`, not from a local run. Say if you want the gate and I'll refresh them that way; otherwise closing both is fine.

iceteaSA:
Landed on main through abe9840 (merge of fd004ad and 3d8a1d1), which the release ratchet in ce0d9f6 builds on. GitHub didn't mark it merged because I force-pushed a rebased head after the merge. Closing.

Issue #171, opened 2026-08-28
packages/core/src/tests/ never runs — 153 tests across 7 files are unreachable from any gate or CI step


iceteaSA:
The seven test files under `packages/core/src/tests/` do not execute in any gate or CI step. They pass when run directly, so nothing looks broken — they just never run.

## Mechanism

The root test script scopes to opencode:

```json
"test": "bun run build && cd packages/opencode && bun run test"
```

`packages/core/package.json` has no `test` script at all:

```json
{ "build": "...", "typecheck": "tsc", "types": "tsc", "prepublishOnly": "bun run build" }
```

And CI (`.github/workflows/ci.yml`) runs only:

```
bun run types · bun run build · smoke:tui · bun run test · bun run test:e2e · format:check · lint
```

`bun run test` reaches `packages/opencode` (42 files). `test:e2e` builds core and then runs `packages/e2e-tests`. Neither path reaches `packages/core/src/tests/`.

## Verified on `db2f6c1` (v1.21.0)

```
$ bun run test
Ran 1312 tests across 42 files.          # opencode only

$ cd packages/core && bun test src/tests/
 153 pass
 0 fail
Ran 153 tests across 7 files. [454.00ms]  # never reached by any gate
```

Stranded files: `dump`, `killswitch`, `models`, `prime-manager`, `prime`, `prime-usage-race`, `quota-surfaces`.

## Why it matters

That is not a marginal set. It covers killswitch threshold logic, prime scheduling and its usage race, dump sweeping, and quota surfaces — behavior that gates spending and blocks requests. A regression in any of it lands green.

The failure is silent in the worst direction: the files exist, they are well-formed, they pass locally when someone runs them directly, and CI is green. Nothing distinguishes "these tests protect this code" from "these tests have not run since they were written." A test that cannot run is not a test, and this shape hides that fact rather than surfacing it.

I found it by chasing a test-count anomaly — a subagent reported `+1` after adding 4 tests to `packages/core/src/tests/`, and the arithmetic did not work out. Without that mismatch I would not have looked.

## Suggested fix

Add a `test` script to `packages/core` and include it in the root gate, e.g.:

```json
// packages/core/package.json
"test": "bun test src/tests/"

// root package.json
"test": "bun run build && bun run --cwd packages/core test && cd packages/opencode && bun run test"
```

[...]

PR #235, opened 2026-09-17
ci: add test-count floor gate with merge-target ratchet

iceteaSA:
Stacked on #234 — merge that first, or this compares against a base with no floors and reports UNCHECKED.

## What this closes

We have no test-count floor at all. The `check` gate is `build && core && opencode && pi`, so a suite can shrink to a single test and stay green.

That is not hypothetical. In August, 153 tests in `packages/core` were stranded — present, correct, and executed by no gate (#171). It was caught because a test total rose by fewer than the tests added and the arithmetic did not work. Nothing mechanical would catch a recurrence.

## Three enforcements, kept distinguishable

```
COUNT:   measured >= branch floor          the suite did not shrink
RATCHET: branch floor >= merge target's    a branch may raise, never silently lower
LOWERING: allowed, with an explicit from/to marker and a reason
```

The ratchet is the part that is easy to get wrong. A bare number bumped by whoever measures is **carryable backward by a stale branch**: a branch forked before a bump merges green and lowers the bound, and nothing goes red, because a floor is a lower bound and lower bounds do not complain about being lowered. So the gate compares against the merge target's value rather than its own.

`fetch-depth: 0` is load-bearing — a shallow clone cannot resolve the merge target's floor, and without it the UNCHECKED arm would be the only arm that ever fired.

## Three verdict states, not two

```
FAIL       a floor was violated, or the invocation was noncompliant
UNCHECKED  the merge target could not be inspected  (exit 2)
PASS
```

Missing **branch** config is a noncompliant source contract → FAIL, rebase and adopt. Inability to inspect the **merge target** → UNCHECKED. Collapsing those would make UNCHECKED a catch-all that quietly absorbs a real violation.

One defect found by asking what `github.event.pull_request.base.sha` does on a non-PR event: `if (!value || …)` treated a quoted `--base-ref ""` as a missing flag and crashed as FAIL/1 rather than UNCHECKED/2. Every direct push would have hard-failed the gate — and failed as a *violation verdict*, asserting floors were breached when the comparison was never attempted. A gate that cries violation when it cannot run gets disabled. The suite was green with that defect present, because no arm passed an empty flag, only a missing one. Now pinned.

## Verdicts carry their scope

```
VERDICT: PASS packages=core,opencode,pi (test counts and floor ratchet satisfied)
VERDICT: UNCHECKED packages=core,opencode,pi (merge target ref was not provided)
VERDICT: FAIL packages=none (test-count floor check rejected invalid input)
```

The package set is derived from the counts actually evaluated, not a hardcoded list. An empty scope renders `packages=none` and never as a pass. A green that does not state its scope can be read as covering more than it does — if a package were dropped from the checked set tomorrow, a scope-free verdict line would look identical.

## Proof arms

```
a  counts equal to floors                 PASS      0
b  counts above floors                    PASS      0
c  a count below its floor                FAIL      1
d  branch floor below merge target        FAIL      1
e  deliberate lowering with marker        PASS      0
f  unresolvable base — missing, empty,
   and absent ref                         UNCHECKED 2
g  stale-branch replay                    FAIL      1
   branch has no floor config             FAIL      1
```

Arm (g) is the one that matters: floors bumped on main, a branch forked before the bump carrying the old numbers. Without it the other arms only prove the script runs.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/anthropic-auth/pr/235"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792248024&installation_model_id=22398&pr_number=235&repository=cortexkit%2Fanthropic-auth&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fanthropic-auth%2Fpull%2F235&signature=8d700159b2ed5a6c497e43d637dc3a29e46b374978c39a9510437f37ba4dd1df"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Adds a test-count floor gate with a merge-target ratchet so a test suite can no longer shrink to a single test and stay green. It records initial floors for `core`, `opencode`, and `pi` and measures counts itself by running each package's tests.

**Behavior**
- Checks each package's measured test count against its floor, and that branch floors are never lower than the merge target's floors.
- Deliberate lowerings are allowed only with an explicit marker file naming the from/to values and a reason.
- Returns PASS, FAIL, or UNCHECKED; missing branch floor config is FAIL, while an unresolvable merge target or a measurement stamp that isn't a clean ancestor of the checked ref is UNCHECKED (exit 2).
- Verdicts include the package scope actually evaluated, so a green result can't be misinterpreted.
- CI now uses `fetch-depth: 0` so the merge target's floor can be resolved.

<sup>Written for commit fd004ad155dd97e76e4be3c9bd10b743134d2516. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/anthropic-auth/pull/235?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


iceteaSA:
CI on this PR is red, and it is red for the right reason. Recording it rather than working around it.

```
VERDICT: UNCHECKED packages=core,opencode,pi
  (could not resolve merge target floor f9c867d1…:.ci/test-count-floors.json)
exit 2
```

GitHub sets `base.sha` from the PR's **base branch**, not from its stacked parent — so this PR's merge target is `main` at `f9c867d1`, which does not yet contain `.ci/test-count-floors.json` (verified: 404 on that path at `main`). The gate cannot resolve a floor to compare against, so it reports UNCHECKED and exits 2.

That is the gate doing its job. It refuses to report PASS when the comparison was never attempted, which is the property the whole change exists to establish.

**#234 merges first, then this goes green on a re-run.** No code change is needed here and none should be made — the fix for "the gate cannot compare" is to give it something to compare against, not to teach it to pass anyway.

Worth stating plainly because it is the honest correction to my own PR description: I wrote that splitting the landing means "the base already has floors, so the ratchet works on its first run". That is true only *after* #234 lands. Until then this PR is correctly blocked, and the split still does what it was for — there is no bootstrap exemption anywhere in the code, and the condition resolves by merging in order rather than by an authorized bypass.

iceteaSA:
Marked draft until #234 lands, because a red check here reads as "broken" when the accurate signal is "not mergeable yet".

The red is the gate refusing to report PASS on a comparison it could not make — `base.sha` is `main`, which has no `.ci/test-count-floors.json` until #234 merges. Nothing to fix here; it cannot go green before then, and it should not.

#234 is green and independent. Once it merges I will re-run this and un-draft.

iceteaSA:
@ualtinok, following up. Rebased onto the rebased #234; it still fails here as expected (`UNCHECKED`, exit 2) because `main` has no floors file until #234 merges.

One thing I found by running it against current `main` that's worth knowing before you decide. The script **refuses to count a test run that fails**:

```
opencode test command exited 1
VERDICT: FAIL packages=none (test-count floor check rejected invalid input)
```

That's deliberate and I think right: a count from a failing run can't be trusted. The side effect is that the floor gate fails whenever the test gate fails. So it doesn't add flakiness beyond what #220 already causes, but those flakes will show up twice. On a green run, passes equal tests run, so counting passes rather than tests run makes no practical difference.

What it catches that nothing else does: a suite that **silently shrinks**. For example, when the core tests were never being run from the root gate (#171), 153 tests were skipped and CI stayed green.

iceteaSA:
Landed on main through abe9840 together with #234. The release-tag comparison in ce0d9f6 extends it. Closing.

PR #236, opened 2026-09-18
fix(core): reject empty segments in custody credential IDs

iceteaSA:
`isScopedCustodyCredentialId` checked that segment 1 matched the provider but never checked the segments were non-empty, so `:anthropic:x`, `oauth::x`, `oauth:anthropic:` and `oauth:anthropic::y` all passed.

The handle manifest is a **co-tenant file** — a sibling plugin writes its own block into the same file. Claustrum has tightened their parser to reject any empty segment, and **their parser refuses the whole file on a malformed row while ours isolates the bad row to `corruptLabels`**. So the writer was the real exposure: emitting `oauth:anthropic:` hands a co-tenant a row that takes their entire manifest down, and their failure mode is total rather than per-row.

## The rule

Every colon-separated segment must be non-empty, in addition to `segments[1] === provider`.

The kind segment stays an **open set** — deliberately not enumerated. Four kinds are live in the vault today (`oauth:`, `chatgpt:`, `antigravity:`, `apikey:`) and enumerating them would reject a new one the moment it ships. The label (segment 3+) is still never consulted here; it is a lookup key elsewhere, not an authorization check.

## Remover left deliberately weak

`removeCustodyHandleManifestEntry` keeps the permissive `isValidCustodyCredentialId`. Tightening it would **strand** an already-written malformed entry — the entry could never be removed by the code that refuses to name it. Removal is a cleanup path and must stay able to address anything the writer previously let through.

## Verification

Independent mutation proof per site, because reader and writer call the same helper and one test can go green while the other site is untested:

| reverted site | test that goes red | result |
|---|---|---|
| writer only | `writer refuses empty-segment credential ids before they reach disk` | expected refused, got `{ status: "written" }` |
| reader only | `rejects credential ids with any empty colon-separated segment` | expected `Set {"empty-kind"}`, got `Set {}` |

Conformance probed against the built predicate, not only via unit tests — the accept rows are live vault credential shapes and a regression there would mean the kind segment had been narrowed:

```
ok   "oauth:anthropic"            provider=anthropic   accepted
ok   "oauth:anthropic:work-alt"   provider=anthropic   accepted
ok   "chatgpt:openai"             provider=openai      accepted
ok   "antigravity:google"         provider=google      accepted
ok   "apikey:deepseek:main"       provider=deepseek    accepted
ok   ":anthropic:x"               provider=anthropic   refused
ok   "oauth::x"                   provider=anthropic   refused
ok   "oauth:anthropic:"           provider=anthropic   refused
ok   "oauth:anthropic::y"         provider=anthropic   refused
ok   "oauth:openai"               provider=anthropic   refused   (cross-provider, unchanged)
```

Gates: `packages/core` 202 pass / 0 fail (488 assertions), root typecheck clean.

Context: cortexkit/claustrum#44 (the rule) and #45 (per-row refusal as its safety precondition).

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/anthropic-auth/pr/236"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792312612&installation_model_id=22398&pr_number=236&repository=cortexkit%2Fanthropic-auth&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fanthropic-auth%2Fpull%2F236&signature=10ae9cc3a2364c14d5c4e316756b127184086754adc7ebf4629a9dbcaf29c78e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Rejects empty colon-separated segments in custody credential IDs so malformed IDs can’t reach the shared manifest, where a co-tenant parser rejects the whole file. Previously only the provider segment was checked, so IDs like `oauth:anthropic:` were accepted on write and treated as valid on read.

- Every segment must now be non-empty; the kind segment stays open to future values and the provider segment must still match.
- Removal stays permissive so malformed entries written before this change can still be deleted.
- Tests cover reader reporting, writer refusal, and not creating a manifest file on refusal.

<sup>Written for commit 6a2e7dab06af768286fc6e13b5ea38fa99a66a3b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/anthropic-auth/pull/236?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


iceteaSA:
@ualtinok, following up: re-checked against current `main` (`4a01612`) and this is still needed. `packages/core/src/claustrum.ts:412-419` still checks only the provider segment:

```ts
const segments = value.split(':')
return segments[1] === provider
```

so credential IDs with an empty segment still pass (`:anthropic:x`, `oauth:anthropic:`). Claustrum now rejects any empty segment on its side (cortexkit/claustrum#44, `4224e8c`). As a result, our reader can accept, and our writer can write back, an ID that Claustrum's parser rejects for the whole file. A shared handle manifest is exactly where that difference causes damage.

Rebased with no conflicts; CI is green.

iceteaSA:
No longer applies: c2f3428 removed manifest handle resolution, so `isScopedCustodyCredentialId` and `readCustodyHandles` no longer exist on main. Closing.

PR #238, opened 2026-09-18
fix(opencode): accept effort-marked Fable 5.1 tool continuations

### LINKED ISSUE #237: Fable 5.1 effort-marker correlation fails when a turn has more than one user message
## Symptom

A Fable 5.1 turn dies with a user-visible 400 carrying an internal message:

```
Missing or invalid internal Fable 5.1 effort anchor
```

Recurring, not a one-off — **8 occurrences across 3 sessions in one week**, in four shapes, all `EffortMarkerCorrelationError`:

| count | message |
|---|---|
| 4 | `Fable 5.1 effort marker correlation failed: expected N, found M` |
| 2 | `Multiple internal Fable 5.1 effort markers on one user boundary` |
| 2 | `Missing or invalid internal Fable 5.1 effort anchor` |

## Root cause

`buildEffortRequestPlan` pushes the anchor onto the user message that is current **when the messages hook runs** (`effort-history.ts:409`, `currentUser.parts.push(...)`), and stamps that message's id as `anchor.boundary`.

Validation at lowering time then requires the anchor to sit on the **last** user message:

```ts
// effort-history.ts:636-646
if (
  requestPlan.markerCount > 0 &&
  (anchors.length !== 1 ||
    anchorMessageIndex !== lastUserMessageIndex ||   // <-- this
    !anchor ||
    anchor.scope !== requestPlan.scope)
) throw new EffortMarkerCorrelationError('Missing or invalid internal Fable 5.1 effort anchor')
```

**The invariant assumed is "exactly one user message at the current boundary."** That holds for a plain chat turn and breaks as soon as the host puts a second user message at the same boundary — after which the anchor is on an earlier user message and the check fails.

## Evidence

From the failing session (ids abbreviated):

```
13:48:37  user       "Do a work laptop sync, ORW as well"      <- operator prompt
13:48:37  user       <external-context source="sibling-session" …>   <- second user message, same second
13:48:37  assistant  ok
13:49:10  assistant  ERR   <- tool continuation, no new user message
13:49:43  user       <task id="…" state="completed">           <- another injected user message
13:49:43  assistant  ok
13:49:58  assistant  ERR   <- tool continuation
```

Both failures are the **second provider call within a turn** (tool continuation), in a turn whose boundary carries two user messages. Neither injected message is flagged `synthetic` or `ignored` — they are ordinary user messages as far as the plugin can see.

The `Multiple internal Fable 5.1 effort markers on one user boundary` variant is the same root cause seen from the other side: more than one user message at a boundary means more than one marker-bearing candidate.

## Why it matters

The fail-closed itself is behaving as designed — refusing before cache shaping, signing and transport is the right call versus shipping a bad prefix. Two things make it worth fixing anyway:

1. **The trigger is benign.** A trailing user message carrying no marker does not corrupt the effort timeline the plan describes, so the refusal costs a turn for nothing.
2. **The failure is diagnostically dark.** `effortMarkerFailureResponse` (`index.ts:322`) synthesises a 400 for the user, and the throw happens before the dump path, so there is **no log line and no dump artifact**. Reconstructing the above required the host database. At minimum this path should log the mismatch it refused on (anchor boundary, last user id, marker count) — a fail-closed nobody can diagnose gets worked around instead of fixed.

## Not fixed by f9c867d

`f9c867d` bounds and orders *desktop notice* identities. These injected user messages are not desktop notices, and the strict check is still present on current `main` (`effort-history.ts:639`).

## Suggested direction

Tolerate user messages appended after the plan was built, as long as they carry no markers — i.e. require the anchor to be on the last user message **the plan knew about**, rather than the last user message in the lowered history. Holes, reordering, mutation, and scope/digest mismatch should keep failing exactly as they do now.

Environment: this is heaviest in multi-seat setups where session-to-session frames and subagent completion frames are delivered as ordinary user messages, but any host that appends a second user message at a turn boundary will hit it.

iceteaSA:
Closes #237.

A Fable 5.1 turn dies with a synthetic 400 carrying `Missing or invalid internal Fable 5.1 effort anchor` whenever the model makes a tool call. Seven occurrences across three sessions in a week, in four message variants.

## Root cause

**Not what #237 guessed.** That issue blamed boundaries carrying more than one user message. The plan lookup was never the problem — OpenCode keeps `chat.headers.message.id` bound to the originating user message throughout a tool loop, so the tracker resolves correctly on every continuation.

The anchor is pushed onto the user message that is current when the messages hook runs. Provider lowering then appends `assistant tool_use` → `user tool_result` after it, because **a tool result is a user-role message in Anthropic's wire format**. Validation required the anchor to sit on the *last* user message:

```ts
anchorMessageIndex !== lastUserMessageIndex   // → throw
```

Reproduced values at the throw:

```
anchorMessageIndex   2          lastUserMessageIndex  4
anchorsFound         1          markerCount           1
anchor boundary      msg_current   planned boundary   msg_current
anchor token == expected token · scopes identical
```

Everything correlated. Only the position differed, and it differed for a reason the check did not model.

## Fix

An anchor may precede trailing user messages **only** when every one of them is a non-empty `tool_result`-only message whose `tool_use_id` appears in the immediately preceding assistant's `tool_use` blocks.

Still fail-closed, unchanged: plain user suffixes, holes, reorderings, mutations, scope mismatch, digest mismatch, anchor-token mismatch, multiple anchors.

## Diagnosability

Equally important, and the reason this took two commits. The two throw sites emitted the **identical string**, and the throw happens before the dump path, so a production occurrence left **zero trace** in the log — all seven had to be reconstructed from the host database.

- distinct checks: `anchor_placement` and `anchor_token`, with distinct messages
- `effortMarkerFailureResponse` logs at warn with check code, anchor and planned boundary IDs, both indexes, anchors found, marker count, scopes, and match state
- expected/found anchors are logged as **SHA-256 hashes**, never raw marker tokens

## Verification

Reproduction: `accepts authenticated effort history on a tool continuation` — pre-fix `0 pass, 1 fail` with `anchor_placement`.

The second commit exists because the first one's safety half was **untested**. Mutating the whole `tool_result` pairing predicate to `return true` passed the entire suite at 1897/0 — the existing rejection test builds `user · user(anchor) · user(plain)` with no assistant between, so it bails at an earlier guard and never reaches the pairing. The uncovered shape is an ordinary new user turn after an assistant reply.

Three tests now pin it, each independently mutation-proven:

| test | mutation that reddens it |
|---|---|
| `rejects a plain user turn after an assistant reply` | whole pairing predicate → `return true` |
| `rejects a tool result that does not match the preceding tool use` | drop `toolUseIds.has(block.tool_use_id)` |
| `rejects a tool-result user message mixed with text` | `.every(...)` → `.some(...)` |

Re-ran the original surviving mutation against the final tree: now **3 fail** where it previously passed clean.

Gates: core 199/0 · opencode 1900/0 · pi 114/0 · typecheck clean.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/anthropic-auth/pr/238?ref=codesmith_pr_footer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792348239&installation_model_id=22398&pr_number=238&ref=codesmith_pr_footer&repository=cortexkit%2Fanthropic-auth&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fanthropic-auth%2Fpull%2F238&signature=d204280b1c5877f2a9ddb71bc0af4174d20c543b5d265a29e44703a3e8826955"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes Fable 5.1 tool calls dying with a 400 error when the assistant makes a tool call. The anchor placement check was too strict — it required the effort anchor to sit on the last user message, but Anthropic's wire format appends tool results as user-role messages after it.

**Bug Fixes**

- Now accepts trailing user messages after the anchor when every one is a `tool_result`-only message whose `tool_use_id` matches a `tool_use` block in the preceding assistant message.
- Plain user suffixes, mismatched or mixed `tool_result` content, scope mismatches, and anchor token mismatches still fail closed.

**Diagnosability**

- The two throw sites previously emitted the same string with no log output, leaving zero trace in production; they now use distinct `anchor_placement` and `anchor_token` check codes with distinct messages.
- Refusals now log at warn with check code, boundary IDs, indexes, and match state — marker tokens are logged as SHA-256 hashes, never raw values.

Closes #237.

<sup>Written for commit 3251e64666fdd0d386c281024f1c07f3eee5bd26. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/anthropic-auth/pull/238?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


iceteaSA:
Rebased onto `main` at `4a01612`. CI is red but **not on this PR's change** — and not on a test failure at all.

The latest run reports `1932 pass, 0 fail, 1 error`. The error is a crash in production code between tests:

```
# Unhandled error between tests
TypeError: undefined is not an object (evaluating 'filename.toString')
  at filenameMatches (packages/opencode/src/custody-manifest-watcher.ts:51:31)
  at <anonymous> (packages/opencode/src/custody-manifest-watcher.ts:142:16)
  at #onEvent (node:fs:49:70)
```

`filenameMatches` guards `filename === null` but Node's `fs.watch` can invoke the listener with `undefined`. The declared type is `string | Buffer | null`, so TypeScript never flags it. Reproduced against clean `upstream/main`; this PR does not touch custody. Fix is on `fix/watcher-undefined-filename` and I will open it separately.

An earlier run on this branch failed differently — on the Claustrum legacy-handle migration tests. Those are load-sensitive and I reproduced them on **clean `upstream/main` under CPU pressure** with no branch applied (2 fail under 16 hogs, 2 pass on the same tree unloaded). Filed on #220.

So this branch has now shown two distinct red runs, neither caused by it: one load-sensitive flake, one genuine upstream crash. Local gates on the rebased branch are clean apart from the known pre-existing failures.

iceteaSA:
CI red again, same cause, now confirmed reproducible rather than intermittent: `1932 pass, 0 fail, 1 error` with

```
TypeError: undefined is not an object (evaluating 'filename.toString')
```

Fix is up as #246 against `main`. This branch is unblocked the moment that lands — nothing here needs changing.

Worth noting the two runs together, because they say different things. The earlier red run was the load-sensitive Claustrum migration tests ([#220](https://github.com/cortexkit/anthropic-auth/issues/220)) — genuinely intermittent. This one has now reproduced twice and is not luck: the `fs.watch` callback fires with an undefined filename on this runner reliably enough to crash every run.

That also means the gate is currently only *probabilistically* green for everyone. #246 passed CI, and it contains the fix; the PRs that went green today did so on runs where the watch event happened not to land inside the test window.

PR #241, opened 2026-09-18
fix(fable): accept effort markers on a merged user boundary

iceteaSA:
Fixes the `Multiple internal Fable 5.1 effort markers on one user boundary` variant of #237. Live failure — four occurrences, each a synthetic HTTP 400 that kills the turn:

```
2026-09-15 09:29:50 · 2026-09-16 18:06:37 · 2026-09-18 20:43:06 · 2026-09-18 20:44:04
```

The last two are on a build carrying #238, which fixed the *other* variant (`anchor_placement`). Separate mechanism, same file.

## Root cause: two sides disagree about what a boundary is

Markers are **attached per host record** — `effort-history.ts:391`:

```ts
const token = transitionMarker(scope, info.id, effort)
item.parts.push({ type: 'text', text: token })
```

Markers are **validated per wire message** — `effort-history.ts:611`:

```ts
for (const message of consumed.messages) {
  if (message.transitions.length > 1) { throw … }
```

`consumeInternalMarkers` groups by wire message. Anthropic requires strictly alternating roles, so consecutive host user records collapse into one wire user message and both markers land on a single boundary. The invariant is broken by construction.

### Why it clusters

Measured from the host DB, not inferred:

- **26 consecutive-user-record pairs** in the affected session. Coordinator seats receive peer session frames and subagent completions as ordinary user-role messages.
- **A failed turn leaves an assistant record with zero parts** — `parts=0, textparts=0` for every errored message id checked. A contentless assistant lowers to nothing, so it cannot separate the user records on either side of it.

So each failure widens the next request's merge:

```
09:29:08  user low
09:29:08  user low
09:29:08  assistant ERR   ← "correlation failed"        (0 parts)
09:29:50  user high
09:29:50  assistant ERR   ← "Multiple internal markers" (0 parts)
```

Self-amplifying, which is why every occurrence sits in a run rather than alone.

## The change

**1. Dropped the per-message count guard.** Duplication and ordering are already enforced globally on the flat array: the count check (`:690`), the ordered `planDigest` (`:695-706`), and positional token comparison in the trimmed path (`:666-675`). The per-message check added no protection those lack — it encoded a structural assumption the lowering violates.

**2. Scope-check every transition on a boundary, not just `[0]`.** With N>1 permitted, a foreign-scope marker at index 1 previously bypassed the explicit check.

**3. Emission takes the last transition, not the first** — `:724`, `transitions[0]` → `transitions.at(-1)`.

Point 3 is the one that makes this safe. A marker means *"effort became X at this boundary"*; when two records merge, the effort going into that wire message is the last one's. **Relaxing the guard without this would turn a loud 400 into a silent wrong-effort request** — strictly worse than the crash it replaces.

## Verification

Three mutations, re-run independently of the implementer's report, tree restored to clean between each:

| mutation | result |
|---|---|
| restore the `> 1` throw | `accepts a merged boundary whose transitions are a correctly-ordered run and emits the last effort` — throws `Multiple internal Fable 5.1 effort markers on one user boundary` |
| revert emission to `transitions[0]` | same test — `effort: "high"` received where `"max"` expected |
| scope-check only `[0]` | `rejects a foreign-scope marker at index 1 of a merged boundary` — `Expected substring: "…scope mismatch"` / `Received: "…non-prefix loss"` |

The second is the important one: it proves the test asserts the emitted **effort value**, not merely that nothing threw. Without that assertion the `[0]` → `at(-1)` fix would be untested and the silent-wrong-value path would ship green.

**On the third — the explicit scope check is defense-in-depth, not the only line.** With it narrowed back to `[0]`, the foreign-scope marker at index 1 is still *caught*, by the flat check, as `non-prefix loss`. The mutation changes which error fires, not whether it fires. Worth stating plainly rather than overclaiming: the check buys a precise diagnostic, not a closed hole.

Gates: **core 199/0 · opencode 1898/0 · pi 114/0** · typecheck clean.

## Scope

Untouched: the anchor logic (`:623-656`; anchors are placed on the current boundary only and re-stripped every request, so merging cannot produce two), the attach site (`:362-395`), and every flat check. Holes, reordering, mutation, duplication, and scope mismatch all stay fail-closed.

One existing test in `index.test.ts` asserted the removed guard's message for a duplicated marker. It now includes the anchor so it exercises the flat count check, and expects `correlation failed: expected 1, found 2`. Duplication still fails closed — 400, request never sent.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/anthropic-auth/pr/241?ref=codesmith_pr_footer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792361870&installation_model_id=22398&pr_number=241&ref=codesmith_pr_footer&repository=cortexkit%2Fanthropic-auth&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fanthropic-auth%2Fpull%2F241&signature=e14c1067ec01e8bc579f42a64df7d35cc661cff7e344141c2c663d81997dd677"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes the crash when multiple effort markers land on one merged user boundary, so those requests no longer die with a synthetic HTTP 400. Consecutive host user records collapse into one wire message, so a boundary legitimately carries several markers; the old per-message guard rejected that shape. A failed turn leaves a contentless assistant record that cannot separate the user records around it, which is why the fault self-amplified.

**Key changes**
- Dropped the per-message count guard; the flat checks already pin ordering, duplication, and position across all messages.
- Scope-check now runs on every transition on a boundary, not just the first.
- Emission uses the last transition (`transitions.at(-1)`) so a merged boundary applies the correct effort value instead of a stale one.

Duplication, reordering, and scope mismatch still fail closed with a 400 before the request is sent. The existing duplicate-marker test now exercises the flat count check and expects the correlation error message.

<sup>Written for commit 72618a7b9a9b8f699995e57a04c03d4901172fa1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/anthropic-auth/pull/241?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


iceteaSA:
Rebased onto `main` at v1.23.0. One conflict, in an import line only: v1.23.0 added `getOpenCodeClaustrumEnrollmentPaths` to `index.test.ts` beside this PR's `EFFORT_ANCHOR_PREFIX`. Resolved as the union of both; no test or production hunk changed.

**Still valid — checked, not assumed.** `packages/opencode/src/effort-history.ts` is byte-identical between this PR's base and v1.23.0, so nothing upstream touched the defect or the fix:

```
1b27511        emission-loop transitions[0] = 2
upstream/main  emission-loop transitions[0] = 2
```

A caution on that check: grepping for `transitions.at(-1)` returns 2 hits on `main`, which reads like "already fixed". Both are pre-existing sites (`:591`, `:707`) that predate this PR. A count without a baseline is not evidence — the base has the same 2. The emission loop this PR changes is still `transitions[0]`.

## Pre-existing failures are upstream's, not this PR's

Two failures on the full opencode suite, both reproducing on **clean `upstream/main`** with none of this PR's changes:

- `vault reauth leaves the account absent and projects the reauth state` — passes alone, fails in-suite (pollution)
- `starts a durable enrollment ceremony in Claustrum mode without enabling scoped serving` — load-sensitive; failed one clean-`main` run, passed the next

Same failing set with and without this branch; counts vary 2→1 across runs on both. Matches #220.

Typecheck passes; core 295/0 on the new baseline.

Issue #230, opened 2026-09-16
Two review findings lost when #218/#229 were closed still apply to main


iceteaSA:
When #218 and #229 were integrated in modified form and closed, their unresolved review threads went with them. Two of the findings apply to the code that actually shipped. Sweeping them here rather than leaving them attached to closed PRs.

## 1. A second desktop notice is not tracked as plugin-generated (billing risk)

`sendIgnoredMessage` sets the tracked id only when the placement helper returns one:

```ts
if (messageID) request.body.messageID = messageID
…
if (request.body.messageID) { options.onPreparedMessageId?.(request.body.messageID) }
```

`notificationMessageIdBeforeAssistant` returns `undefined` when its candidate would sort at or below the latest user message. After notice 1 is inserted it *is* the latest user message, so a second notice against the same active assistant gets no id, the host assigns an untracked one, and `isDesktopNoticeMessage` returns false for it. It is then treated as a genuine user message — which revokes the delivery lease and is the same path that produced a duplicate billed provider turn in #162.

Reproduced against the helper's real logic:

```
notice 1 id : msg_0000000effffzzzzzzzzzzzzzz
notice 2 id : undefined      <- host assigns an untracked id
```

The switch→restoration pair is exactly a two-notice sequence, so this is reachable on the path the recovery fix serves. Original finding: cubic on #229 (`index.ts:657`), confidence 8.

## 2. `desktopNoticeMessageIds` is unbounded in session count

Every sibling in that family is bounded two ways — per-entry and across sessions:

```
pendingDesktopNotices        size > 128 cap ✓
desktopNoticeSafeSessions    size > 128 cap ✓
desktopNoticeMessageIds      per-set cap of 4, no map-level cap ✗
```

The only eviction is `session.deleted`. Sessions that are never deleted — the normal case — leave an entry per session for the process lifetime. Small per entry (a Set of ≤4 short strings), so this is a slow leak rather than a bug, but it is an inconsistency with its siblings in a process designed to run for weeks. Original finding: cubic on #218 (`index.ts:5320`), confidence 8.

Happy to send a PR for either; #2 is a four-line change matching the sibling pattern, #1 needs a decision on whether to mint a unique ordered id per notice or correlate the host-assigned one.

ualtinok:
[...]

PR #242, opened 2026-09-18
fix(fable): make the effort-marker trim tolerance reachable

iceteaSA:
Fixes the `correlation failed: expected N, found M` variant of #237 — the **initiating** failure. It produces a `parts=0` contentless assistant, which lowers to nothing on the next request and so cannot separate the user records around it, widening the next merge into the variant #241 fixes. #241 stopped the amplification; this stops the ignition.

All four fleet-wide occurrences, every one `found < expected` — **loss, never duplication**:

```
2026-09-12 08:47:23  ses_f7ed4e49…  expected 1, found 0
2026-09-14 09:49:53  ses_019303cc…  expected 4, found 1
2026-09-14 17:48:54  ses_f7ed4e49…  expected 1, found 0
2026-09-15 09:29:08  ses_019303cc…  expected 2, found 1
```

Loss already has correct tolerance at `effort-history.ts:658-688`: the host trims a history prefix, survivors must be an exact suffix of the plan, consumed transitions fold into the baseline. **Neither occurring shape reached it**, for two unrelated reasons.

## Shape A — total trim short-circuits the tolerance

`effort-history.ts:570-576` threw unconditionally on `expectedTransitions`, ~90 lines before the tolerance. The result was an asymmetry with no justification: trim 3 of 4 markers and the suffix logic accepts it; trim the 4th as well and the *same legitimate compaction* becomes a 400.

A full trim is simply `trimmedPrefix === expectedTransitions.length` — every transition consumed. It now folds all of them into the baseline. The throw is retained for the genuinely untrusted case: `markerCount > 0` with no resolvable plan.

## Shape B — the plan an in-flight header referenced was overwritten

The tracker held one slot per `(sessionId, messageId)`. When the host trims the history prefix between provider calls, the same current user message is re-marked with a shorter timeline and `record()` overwrites the slot — while a header generated from the previous plan is still in flight. `resolveHeader` re-encodes tracked plans and string-compares, so it matched nothing, `expectedTransitions` was null, and the exact path made any loss fatal.

Fixed with a bounded identity-keyed history map alongside the existing slot map, so a header can still resolve the plan it was minted from.

**The mechanism was demonstrated, not assumed.** Two competing candidates were ruled out with evidence: LRU eviction (the in-flight plan is the newest; `record` + `markHeaders` refresh recency) and `clear()` on the same message id (the transform returns a plan, not null, for a single Fable user). A probe also corrected the original hypothesis — a *growing* tool loop alone leaves the plan identical; the **trim** is what re-records it.

## Verification

Five mutations, re-run independently of the implementer's report, tree restored to clean between each:

| mutation | result |
|---|---|
| restore the unconditional `!hasCandidate` throw | `folds a full trim into the last consumed transition effort when the plan resolves` — `expected 2, found 0` |
| fold to `requestPlan.baseline` instead of the last consumed transition | same test — `effort: "low"` where `"high"` expected |
| `resolveHeader` back to scanning the single-slot map | `resolves a plan header after the same message is re-recorded with a trimmed timeline` — `Expected -13 / Received +1` |
| **disable history eviction** (`while (false)`) | `evicts the oldest plan history entry once the cap is exceeded` — `Expected: 4096, Received: 4097` |
| **evict the newest instead of the oldest** | same test reddens |

The last two were not in the original brief. The fourth caught a real gap: the first implementation's `history` bound was correct but **entirely unasserted** — disabling eviction left all 16 tests green. That map is keyed by encoded plan, so it gains an entry per *distinct plan* — per trim per message — growing faster than the slot map beside it, in a long-lived process. Issue #230 is open on exactly that shape, and introducing a second instance of it inside this fix would have been a poor trade.

The fifth separates *bounded* from *bounded correctly*: evicting the newest keeps `size` at the cap, so a size-only assertion would pass while silently reintroducing Shape B — the newest plan is precisely the one an in-flight header needs. Both halves are asserted.

Gates: **core 199/0 · opencode 1898/0 · pi 114/0** · typecheck clean.

## Scope

Untouched: the suffix, digest, and scope checks (`:665-686`, `:695-706`). Non-suffix loss, reordering, mutation, and scope mismatch all stay fail-closed — this makes a **designed** tolerance reachable rather than adding a new one.

Two existing tests encoded the old throw (the unit full-trim assertion and the integration `missingAllMarkers` case) and now assert the folded success instead.

Stacking note: this branch is off `main` and deliberately avoids `:611-628` and `:724`, which #241 changes, so the two do not conflict.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/anthropic-auth/pr/242?ref=codesmith_pr_footer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792365322&installation_model_id=22398&pr_number=242&ref=codesmith_pr_footer&repository=cortexkit%2Fanthropic-auth&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fanthropic-auth%2Fpull%2F242&signature=471fa90609046d5fe358ea94240643a007d75fdbf979b4f88411c72ad0c22c11"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Makes the effort-marker trim tolerance reachable so legitimate host prefix trims no longer fail as 400 errors. Full trims with a resolvable plan now fold into the last consumed transition's baseline, and in-flight headers can still resolve a plan the tracker has since overwritten.

**Bug Fixes**
- A full trim threw unconditionally before the tolerance; it now folds all consumed transitions into the baseline and proceeds unless no plan resolves.
- The tracker kept one slot per `(sessionId, messageId)`, so re-marking the same message with a trimmed timeline orphaned the header already in flight; an identity-keyed history map now preserves those plans.
- Both maps are bounded, and eviction is asserted — disabling it leaves 16 tests green, so the cap is pinned by test.
- Non-prefix loss, reordering, mutation, and scope mismatch still fail closed.
- Two existing tests encoded the old throw and now assert the folded success instead.

<sup>Written for commit b991d3ccf1d7f27e8b7ff888fdd2e5aa60d97811. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/anthropic-auth/pull/242?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


iceteaSA:
Rebased onto `main` at v1.23.0, no conflicts.

**Still valid.** `effort-history.ts` is byte-identical between this PR's base and v1.23.0 — the `!hasCandidate` unconditional throw is untouched, and nothing upstream introduced the history map this PR adds:

```
MAX_TRACKED_EFFORT_PLAN_HISTORY on upstream/main : 0
```

Worth noting against v1.23.0's changelog, which says it now folds "effort changes removed by downstream prefix compaction into the retained baseline through a checksum-bound current-boundary anchor". That landed in **1.22.0** and is the tolerance at `:658-688` this PR is about — it is the mechanism whose two unreachable paths this PR fixes, not a fix for them. Shape A still short-circuits it ~90 lines earlier on a total trim; Shape B still cannot reach it when the tracker's single slot per `(sessionId, messageId)` has been overwritten by a re-record.

## Pre-existing failures are upstream's

Two failures on the full opencode suite, both reproducing on **clean `upstream/main`** with none of this PR's changes: `vault reauth leaves the account absent and projects the reauth state` (passes alone, fails in-suite) and `starts a durable enrollment ceremony in Claustrum mode without enabling scoped serving` (load-sensitive; failed one clean-`main` run, passed the next). Identical failing set with and without this branch. Matches #220.

Typecheck passes; core 295/0.

iceteaSA:
Superseded. 848d79d added the bounded plan history (`MAX_TRACKED_EFFORT_PLAN_HISTORY`) that fixes the plan-overwrite case. a052a43 went the other way on full marker loss: it's treated as unprovable and diagnosed, not folded into the baseline. Closing.

